PR #4, opened 2021-12-13
add unit test for zero balance at unlock time

redlarva:
add unit test for zero balance at unlock time


nightowl121 (approved):
Fixes the issue in iterator of checkpoint

PR #6, opened 2021-12-26
Bbaln test cleanup

AntonAndell:
@nightowl121 Hey, wanted to get a bit more acquainted with java score and its testing myself, also better knowledge of the boostedBaln contracts. To do this is rewrote some tests a bit.

Ill make this PR in case you want to use my changes to your code. otherwise just reject it if you want to keep it as is. Would also appreciate some feedback on it if you have any.


nightowl121 (approved):
Thanks @AntonAndell for the contribution. These are really nice suggestions to increase the code readability.

PR #7, opened 2021-12-28
add unit test for voting power for boosted BALN

redlarva:


token-contracts/bBaln/src/main/java/network/balanced/score/tokens/BoostedBaln.java
```diff
@@ -194,7 +194,7 @@ private void checkpoint(Address address, LockedBalance oldLocked, LockedBalance
 //      initialLastPoint is used for extrapolation to calculate block number
 //      (approximately, for *At methods) and save them
 //      as we cannot figure that out exactly from inside the contract
-        Point initialLastPoint = lastPoint;
+        Point initialLastPoint = lastPoint.newPoint();
```
nightowl121, line 197: Voting escrow uses lastpoint here.
https://github.com/curvefi/curve-dao-contracts/blob/master/contracts/VotingEscrow.vy#L274

nightowl121 (approved):

PR #9, opened 2022-01-12
added safe math method for calculation

redlarva:


token-contracts/bBaln/src/test/java/network/balanced/score/tokens/StateMachineTest.java
```diff
@@ -126,23 +126,24 @@ public void createLock(Account account, BigInteger value, long unlockTime) {
         votingBalances.put(account, vote);
     }
 
+
```
AntonAndell, line 129: remove

token-contracts/bBaln/src/test/java/network/balanced/score/tokens/StateMachineTest.java
```diff
@@ -126,23 +126,24 @@ public void createLock(Account account, BigInteger value, long unlockTime) {
...
         tokenScore.invoke(account,
-                        "transfer", 
-                        bBalnScore.getAddress(), 
-                        value, 
+                        "transfer",
+                        bBalnScore.getAddress(),
+                        value,
                         increaseAmountParams);
-       
+
```
AntonAndell, line 141: remove

token-contracts/bBaln/src/test/java/network/balanced/score/tokens/StateMachineTest.java
```diff
@@ -158,24 +159,39 @@ public void increaseUnlockTime(Account account, BigInteger unlockTime) {
...
             expectErrorMessage(createLockCall, expectedErrorMessage);
         }
 
+        @DisplayName("balanceOf for future")
+        @Test
+        void invalidBalanceOf() {
+            long deltaBlock = (addWeeksToCurrentTimestamp(lockDuration) - sm.getBlock().getTimestamp()) / BLOCK_TIME + 1;
+            sm.getBlock().increase(deltaBlock);
+
+            Executable actualErrorMessage = () ->
```
AntonAndell, line 187: "actualErrorMessage" kinda makes sense but everywhere else the all called after the actual call. 
The function expectErrorMessage might sound misleading, mabye that was a poor choice by me. If you can think of a better name please change it.

I wanted it to read like "expect Error Message from call equal to error message"
expectErrorMessage(method, errorMessage)

token-contracts/bBaln/src/main/java/network/balanced/score/tokens/utils/MathUtils.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.math.BigInteger;
+
+public class MathUtils {
+
+    public static BigInteger safeSubtract(BigInteger from, BigInteger value) {
+        BigInteger result = from.subtract(value);
+        if (result.signum() == -1) {
+            Context.revert(ErrorCodes.InvalidOperation.getCode(),
+                    "MathUtils.SafeSubtract :: Invalid operation");
```
nightowl121, line 13: I would suggest the error message to be changed to ""MathUtils.SafeSubtract :: Invalid operation for unsigned numbers" since the error is more focused on getting signed number from the subtraction of two numbers which needs to be unsigned.

nightowl121 (changes requested):
Block height, timestamp and supply can't be negative, so I would suggest to implement safesubtract in all the subtraction related to these variables.

nightowl121:
The intended changes from this PR is already included in the boosted baln branch.

PR #11, opened 2022-01-25
Feature/add staked LP contract

nightowl121:
Add Staked LP contract and unit tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,5 @@ fabric.properties
 
 # Ignore Gradle build output directory
 build
+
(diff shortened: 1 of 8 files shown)


gedanziger:
This looks pretty good. I wanted to test in my local alongside the rest of the contracts. Can you help me with the java env setup later?

Code-wise, I agree with the changes you made, but I'll do a test as well.

nightowl121:
* Named in both dex and rewards
* If pool is named, check if it is rewarded in rewards contract

PR #15, opened 2022-01-31
Translate Reserve Fund contract in Java

nightowl121:
* Translate the existing reserve fund contract into Java
* Add new methods to disburse and claim the funds


dbrehmer (approved):
Looks good.

PR #45, opened 2022-02-09
Reserve fund unit tests

simusud:
added Reserve fund test and a dummy loans contract
small changes in Reserve Fund score


nightowl121 (changes requested):
* Use HashMap from scorex library rather than java.util
* Revert on fallback
* Optimize claim method, remove interscore call. We will only distribute sicx and baln token for now.
* Remove redeem fail event log
* Rename package name to include contract name, e.g. `network.balanced.score.core.reserve`

nightowl121 (approved):

Issue #49, opened 2022-02-11
Write a new Batch Disbursement contract


nightowl121:
The contract will claim funds from either daofund or reserve fund. It will distribute those funds to the addresses provided in the contract.

**Required methods:**
- Upload the data (onlyGovernance)
        Disbursement
        recipient : hx..
        List [{Token: , Amount: }, … ]
Details of the address to receive amounts will be uploaded using this method.
- Get those details [address, List[{}], isClaimed] -> Users can check the details of amount receivable
- claim() -> Users will be able to claim through this method
- batchDisburse(Address _source) -> this will claim from the daofund or reservefund

PR #50, opened 2022-02-11
Batch Disbursement contract

nightowl121:
Implementation of Batch disbursement contract

Includes feature of #49


dbrehmer (approved):
Looks good.

PR #51, opened 2022-02-12
Loan contract rewrite

AntonAndell:
Translation is finished. For review ignore integrationtests


core-contracts/loans/src/main/java/network/balanced/score/core/Loans.java
```diff
@@ -0,0 +1,896 @@
...
+import static network.balanced.score.core.Constants.*;
+
+class PrepDelegations {
+    public Address _address;
+    public BigInteger _votes_in_per;
+}
+
+public class Loans {
+    public final BigInteger POINTS = BigInteger.TEN.pow(5);
+    public static final VarDB<Boolean> loansOn = Context.newVarDB("LoanON", Boolean.class);
```
nightowl121, line 37: You need to keep the db keys same as previous python contract.
https://github.com/balancednetwork/balanced-contracts/blob/development/core_contracts/loans/loans/loans.py#L82

core-contracts/loans/src/main/java/network/balanced/score/core/Constants.java
```diff
@@ -0,0 +1,43 @@
+package network.balanced.score.core;
+
+import java.math.BigInteger;
+import score.Address;
+
+public class Constants   {
+    private Constants() {
+    }
+    public static final BigInteger SECOND = BigInteger.TEN.pow(6);
```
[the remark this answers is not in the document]
AntonAndell: aah okey, thinking of having days as longs as well since it would make some comparisons more readable

core-contracts/loans/src/main/java/network/balanced/score/core/Loans.java
AntonAndell: Can fix that, but curious for what reasons, other than keeping the naming as was decided in python

nightowl121: At the end, everything is going to be saved as key, value pair in the blockchain database. We have got the data in the previously used keys, so new variables also should use same key to work for previous variable.

`Context.newVarDB("key name", Value class type)`

AntonAndell: Alright but contracts dont share keys i guess? so can we also update python contracts with java code? 

Then i need to rewrite a couple of things and not use the new object writers i guess. so we have the exact same structure for all the snapshots, positions etc

nightowl121: Yes contracts don't share keys. 

Sorry if you have not been communicated early, Yes we are planning to update the existing python contracts to Java. This will keep the contract address same. This can be done. For this you would need to consider that the constructor will be recalled when we update the python contract, so make sure that the previously used on_install code would not re-run again when we update the python contract to Java. 

Yeah the object writer should not be used to translate the previously existing DBs. We need to have the exact same structure for all the DBs, so that there won't be any requirement for migration.

If you need any help on translating some previously used DBs, let me know. I think translating the sub dbs would be little tricky.

AntonAndell: But if we can update it might make more sense to keep the string database structure. But we should probably move away from it at a later point.

AntonAndell: Yeah ill change that, and move towards a more exact copy of the data structure. Thought we where doing data migration during this translation as well. Since we will have to do it at some point i guess. But it might be easier to do once already in java.

AntonAndell: I could use some tips on translating the linkedListDB since it is typed, not sure how to do it in a nice way. Has any1 else translated the linkedListDB yet?

AntonAndell:
@nightowl121 The rewrite is now done correctly but have a few questions:
"make sure that the previously used on_install code would not re-run again when we update the python contract to Java.": What do you mean by this, and how do you prevent this? 

Have any1 else started translating the LinkedListDB yet? or should i do it now, only thing i am missing to start testing now. 

When you create a subDB is there any delimiter there between the subDB key and its own keys:
Ex if i create a sub db get_sub_db("A|B")
then i create a VarDB("C", subdb). Is the key then A|BC? or is there some builtIn delimiter there?

nightowl121:
* Constructor
The constructor is called every time when the contract is installed or updated. In python contract on_install was called only once when contract was installed and on_update was called everytime when contract is updated.
Since we are updating contracts from python to Java, the constructor code should not execute in mainnet as it has already been called but while deploying the contracts to setup an environment in testnets, we need to run those on_install codes.

For example:
In [IRC2Basic](https://github.com/sink772/javaee-tokens/blob/master/tokens/src/main/java/com/iconloop/score/token/irc2/IRC2Basic.java#L39) contract, the name is checked if it is null. This will ensure that if the name is already set, the code in the if loop would not execute. which is like running on_install code.

* Linked List
Yeah we just started linkedlistDB for staking from today. I think we can do specific version for loans. If the python version of linked list is same as staking then, we can do it as a common library.

* Sub DB
For sub db the following code are similar in python and Java 

Python
```python
_sub = db.get_sub_db(b'<sub>')
_var = VarDB("<name>", _sub, value_type=str)
_var.set("<value>")
```

Java
```java
BranchDB<String, VarDB<String>> _var = Context.newBranchDB("<name>", String.class);
_var.at("<sub>").set("<value>");
```

AntonAndell:
Okey thanks, Is there a way to split it the same way in Java or how do we prevent the constructor from running on mainnet?

Those database does not look very similar, The loans one just seemed very general, but i will create a specific one that will only work for the LoansContract then.

nightowl121:
There is no way to split similar to python. Only way to do is to manually make a check if any value is initialized then not to run the on_install code. Like one example could be if governance is set, not to run on_install code.
Solidity upgradeable contracts has a concept of [initializer](https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable).

Yeah BranchDB is the alternative for Sub DBs.

AntonAndell:
@nightowl121 
Hey, code is soon ready to start test upgrading the python version, do you have any test procedure in place for this? or how do you recommend id go about that. 
Also can you look at the day calculations in the getDay function. Don't know why but it seems to do 2 day steps for some reason, don't know what i am missing.

nightowl121:
Hey @AntonAndell, 

This is one of our major and a big component of Balanced. Till now, we have only done unit testing and manually checked if keys remain same and readonly gives same data. We don't have any exact test procedure for migration.
For this one, I would recommend to introduce integration testing and tests for migration. For migration check in integration testing, we can use the zip files of the contract in the development branch in python repo. Following are the steps that can be followed:
- Add all the zip files of python contracts in this repo.
- Deploy those python contract zip files.
- Run the integration tests against python contract
- Upgrade to java contract
- Run the integration tests again.

Sure, will check all the calculations code.

AntonAndell:
Alright that sounds like a pretty good idea, probably gonna prioritize that before continuing with unit tests then. Have similar to this done in any other repo yet that you know of?

nightowl121:
Yeah there are some examples of doing integration tests.

* https://github.com/icon-project/javaee-annotation-processor/tree/master/example/src/test/java/foundation/icon/score/example
* https://github.com/icon-project/btp/blob/iconloop/javascore/test-lib/src/main/java/foundation/icon/score/test/ScoreIntegrationTest.java
* https://github.com/icon-project/java-score-examples/tree/master/testinteg

AntonAndell:
@nightowl121 Having some trouble getting my contracts to deploy from code, I took the helper classes from the java score examples repo since they seemed like a good start for building integration tests.  
I can run the integartion tests of the examples on my local node, i can also manually deploy the Loans contract with the goloop cli. But when trying to deploy through code i first get:
`estimateStep failed(-30005, SystemError(-30005): E0005:IllegalFormat); use default steps.` 

Then followed by a timeout of the deployment call:
`network.balanced.test.ResultTimeoutException at LoansUpgrade.java:71`

Do you have any1 idea what i might be doing wrong?

AntonAndell:
@nightowl121 I've realised i can not use any mutable data struct pretty much, Most of the usages can be solved quite easily i think but the caches used by the databases will be very hard to replicate in a good way i reckon. Should we just go without them?

nightowl121:
If we can achieve that through other methods, it should be fine. Overall idea is to reduce db read/write such that the transaction cost is minimized.

AntonAndell:
To be honest a lot of the places where we used maps in python is going to be hard to replicate, right now i do not have a good approach to replicate it. Any input would be great. 

For arrays something like this is possible: 
```
String[] deadAssets = new String[activeAssets.size()];
    int nrDeadAssets = 0;
    for (int i = 0; i < activeAssets.size(); i++) {
        String symbol = activeAssets.get(i);
        Asset asset = get(symbol);
        if (asset.isDead()) {
            deadAssets[nrDeadAssets] = symbol;
            nrDeadAssets += 1;
        }
    }
}
 ```
  
But for maps i dont now how to do it since i cant assign a size to them. Instead there seems no way to dynamically create a list with variable keys and values

AntonAndell:
@nightowl121
I have now a structure of the integration testing that i like, and is what is plan to now start using to integration test the contract. But am i the first to implement integration tests? since there are a lot of files added to the main directory i reckon there should be some decision on what to add there soon before we all start bundling our own integration Test libraries.

nightowl121:
Hey, it seems that the zip files are ignored. 

I looked at the code and it seems that you have done a lot of rewrites just to build transaction. You can use score client to simplify the process of writing the integration tests. 
Here are some examples for that:

- Score Client Documentation:
https://github.com/icon-project/javaee-annotation-processor/tree/master/score-client#for-integration-test-in-score-development-project

- Implementation example:
https://github.com/icon-project/javaee-annotation-processor/blob/master/example/src/test/java/foundation/icon/score/example/HelloWorldTest.java

AntonAndell:
Yeah added the zip files now.
Will look into score annotation processor. Should be smoother than implementing the rpc of each method. But it does however hide the rpc structure of each method which is important, especially during this upgrade since so much is changing. 

I have found it quite hard to navigate all the libraries and such that are available especially for testing. I am still looking in to  how we can write good unit tests. So if you got any tips there that would be great. We don't generally want to depend so much on external contracts and internal state.

AntonAndell:
These score interfaces should  mabye be made as a balanced package in the main directory, Since the also can be used in the SCORE  to call external contracts. We probably should create a utility package available to all contracts  eitherway,

AntonAndell:
@nightowl121  Not sure if you will be the one reviewing but moving this into review now so feel free to start whenever/assign some1.

Some notes:
I have not yet done interfaces with the Score client, couldn't figure out how to achive all the same features i have with this setup. But its pretty good now to write integrationtests for upgrade so we can keep it like this until the Loans contract is done. Then create issues for interfacing all out scores. Not sure if we should go with score client though, need to look more into that library.

For IntegrationsTests i  would like to also calculate snapshots as well as do liquidations and retireBadDebts. But is for now only tested in unit test.

I added a Mock Contracts Package, comes in handy for both unit and Integration Tests, We should however put some thought on how to maintain the mocks if we want to keep a structure like this. 

For the Loans contract, the only things that are missing is Caches and delegate method.
Getting caches to work in java is not very easy, if you have any ideas for it we could add that aswell. 
Have not yet figured out how to take arrays as input in java since they are not allowed, do you know?

There are also a few unit tests missing, but most methods with logic are tested.

nightowl121:
Staking contract has implementation of score client for integration test. You can take a look at how it is done.

AntonAndell:
Is this the folder structure we are going with, the one in the Staking contract? It seems a bit unplanned and not very well thought through, I just dont want to rewrite all the integration testing more than once

AntonAndell:
> Staking contract has implementation of score client for integration test. You can take a look at how it is done.

Mabye we should just pause integrationTesting until we got something on master to base it on? For me to do another rewrite of a full balanced deployment with scoreClient would be quite a large task, Could also rebase my branch on the staking one but might become a bit of a mess. Also i feel like there is a large risk that branch will go over more major changes to the testing structure/utils. So once my branch is ready to merge my changes wont merge well into main. We should however strive to get something similar to what i have in my current environment.  

A governance class that can setup a full balanced environment with the possibility of overriding the "mainnet" contracts with the one currently in development. And also have this class hold all the scoreClients so that the can be fetched from there and not instantiated in the test cases. And the of course make them replaceable with mock contracts to ease some constraints of testing . 

This is if we even want "full" integrationTests. Mabye we want it is done in the staking contract where only the necessary contracts are setup for each test suite.  

Anyhow i'd appreciate some thoughts and plans about testing so i just don't mindlessly write a lot of code that never reaches production.

nightowl121:
Yeah I agree with you, the staking contract don't have the proper folder structure in test setup. This will be modified. 

Regarding the setup of entire balanced in testnet, there needs to be some work where there will be a new subproject dedicated to setting up entire balanced. The tests which are more suitable in each score project can be written in intTest of each score. The tests which will be testing the cases for multiple contracts or more like main features of balanced can be in common project. 

For now, I would suggest to pause your writing of integration test. Reviewing the score would take a while and could increase the requirement of some integration test. Score client makes it easy to write the integration tests as well as it would help in creating score interface to make interscore calls.

AntonAndell:
> Regarding the setup of entire balanced in testnet, there needs to be some work where there will be a new subproject dedicated to setting up entire balanced. The tests which are more suitable in each score project can be written in intTest of each score. The tests which will be testing the cases for multiple contracts or more like main features of balanced can be in common project.

I still think it makes somewhat sense to not remove to much of the system in the intTest. Also makes the intTest of the contracts more similar and easier to grasp if they all base themselves on a full deployment. Only problem is they might take a bit longer to run. 
Not sure about using a separate setup for balanced main features/multiple contracts, since there is a chance such tests are not properly maintained. IMO the optimal would to eventually have each integration test of each contract fully test itself in the full context of balanced. That would make the most robust long term solution. "Skipping" features to test in the general testing suite could create a grey area in testing which can cause problems. 

My integration tests right now are very over engineered and wish i started with the scoreClient earlier but still the structure i quite nice. And something very similar should be achievable with scoreClient. 
So as i mentioned something like a Balanced or Governance class that has all the helpers to setup a balanced environment.
All intTest would then just be.
SetupBalanced ->
fetch ScoreClient for contract to be tested ->
run scenarios. 

This gives reproduce-ability across all intTest for all contracts. Making it fast and safe to develop new features on Balanced should be the top prio during this migration. So  its good that we do not rush it and just go with something. Testing is what makes of breaks a product, especially a open source one :smile:

ZeChosenName:
https://github.com/balancednetwork/balanced-contracts/blob/bb59d33079c619638382c510f18dea6f8b0f4c47/core_contracts/loans/loans/loans.py#L231-L265
Why are these two methods missing in java contract?

AntonAndell:
> https://github.com/balancednetwork/balanced-contracts/blob/bb59d33079c619638382c510f18dea6f8b0f4c47/core_contracts/loans/loans/loans.py#L231-L265 Why are these two methods missing in java contract?

You are right data migratios is missing completely, so its slightly more than thoose two methods even. Dunno why though will investigate.

ZeChosenName:
https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L219-L223
For parameter annotated with `@Optional`, the checks should be done with `null` rather than `BigInteger.ZERO`

ZeChosenName:
https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L295-L300

The `_collateral` parameter should be `Optional`

ZeChosenName:
> > https://github.com/balancednetwork/balanced-contracts/blob/bb59d33079c619638382c510f18dea6f8b0f4c47/core_contracts/loans/loans/loans.py#L231-L265 Why are these two methods missing in java contract?
> 
> You are right data migratios is missing completely, so its slightly more than thoose two methods even. Dunno why though will investigate.

Yes, even `getDataCount` method was missing.
https://github.com/balancednetwork/balanced-contracts/blob/bb59d33079c619638382c510f18dea6f8b0f4c47/core_contracts/loans/loans/loans.py#L502-L508

AntonAndell:
@ZeChosenName Think i found it will check exactly when i get home but seems i was on a slightly old commit on the python repo while doing the transaltions so missed the merge of user migration. 
 Should be this commit https://github.com/balancednetwork/balanced-contracts/commit/2a8f61916f4db32d299f0f9959318dcfbf7a809d
 
 and the dataCount ones

ZeChosenName:
https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L930-L933
https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L941-L944
https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L952-L955
https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L963-L966

The address must be checked whether it is a contract address.

ZeChosenName:
https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L376-L390
The function should return `ArrayList` rather than a single `boolean`

AntonAndell:
> https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L376-L390 The function should return `ArrayList` rather than a single `boolean`

 Do you know how the frontend uses that function, the design does not make much sense. Since it just returns a boolean with its input, so just returning a boolean is the same internally

AntonAndell:
> https://github.com/AntonAndell/balanced-java-contracts/blob/821544e2ea6331aad7893c30e0e52b052334bf7b/core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java#L295-L300
> 
> The `_collateral` parameter should be `Optional`

@ZeChosenName 

Hard to do since the original version was optional with base value true, Donno if there is a solution to it so we can figure out if it was left out or not. However i dont think any1 used that parameter as optional.

Pushed a fix to all your comments so far. Please continue to review, but id prefer if you do it as a review so i can respond in you  suggestions

simusud:
@AntonAndell can you please move this branch to our balanced-java-contracts repo ?

AntonAndell:
@simusud For sure, should definitely be moved to this repo.
https://github.com/balancednetwork/balanced-java-contracts/pull/99

Closing this

PR #54, opened 2022-02-15
(Re)adding initial rebalancing changes

addouglas:
Still needs tests but has a few updates to address constructor and removal of a sICX Receivables VarDB entry that was previously only removed on update. If we see a need to add back the removal, please let me know.


addouglas:
I'm trying to figure out the "why" behind the way admin and governance states/properties were setup in the original rebalancing contract and why so few "get" methods were included in the original. Not sure the best place to ask the question - basically it's odd so I made some tweaks to bring it inline with other contracts... but I can match it 1 for 1 if you'd prefer.

In the original python code - on_install is setting *admin* to governance address which is passed in. I confirmed that governance was in fact what was called on the initial creation of the contract. But it sets admin to the governance value passed in and there's no way to set admin. So all the checks in the original for owner are logically actually being checked against governance as far as I can tell. Also, there are essentially no "gets" for any of these state values. It's like we wanted them secret even though other contracts expose them (more or less). Note: I'm not suggesting we change the admin value as part of the migration... just don't understand why it was set to governance and stuck that way.

In migrating, I standardized it in the pull request with other java contracts... so that admin can be set by governance, governance passed in on constructor sets *governance* and all the state items are gettable. But, like I said, not a strict translation though it should work as the original with the only real exception that admin isn't defaulted on install. But now it can be set. (Note: Loans, for example, sets Admin to Owner on install)

Do you want them strictly like the original in terms of gets and on_install? Do we want it standardized but maybe keep the set admin on install too still just for consistency? 

Note: the very first contract call made was to setGovernance since it was not originally initialized :)

nightowl121:
You can add 'gets' as per requirement. There is no secrets to keep in contracts. Even if there are no get methods, analyzing blockchain data will easily provide the db values.

addouglas:
All existing tests migrated and pass. I noticed that there is not currently a negative rebalance and I've not tried to create such a scenario (yet).

AntonAndell (changes requested):

AntonAndell:
> All existing tests migrated and pass. I noticed that there is not currently a negative rebalance and I've not tried to create such a scenario (yet).

Would be nice with test to cover all the cases of re balancing
All 4 paths available here:
   ```
     if (tokenAmount.compareTo(BigInteger.ZERO) == 1) {
            if (higher) {
                Context.call(loansScore, "raisePrice", tokenAmount);
            }
        }
        else {
            if (lower) {
                Context.call(loansScore, "lowerPrice", tokenAmount.abs());
            }
        } 
```

**Comment by AntonAndell** [comment]
@addouglas  If want some help with evaluating the test you can add jacoo test coverage, so you can see what code you cover int the unit tests. you can see it in the loansContract rewrite or on the bnusd pull request i showed how to add it.

File: core-contracts/Rebalancing/src/main/java/network/balanced/score/core/rebalancing/Rebalancing.java [hunk 1]
```diff
@@ -0,0 +1,204 @@
...
+        return bnusd.get();
+    }
+
+    @External(readonly = true)
+    public Address getSicx(){
+        return sicx.get();
+    }
+
+    private BigInteger calculateTokensToSell(BigInteger _price, BigInteger _baseSupply, BigInteger _quoteSupply) {
+        return _price.multiply(_baseSupply).multiply(_quoteSupply).divide(EXA).sqrt().subtract(_baseSupply);
```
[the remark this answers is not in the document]
AntonAndell: I kinda get the math now but dont really know a good way to split it

AntonAndell:
@nightowl121 You taking over this review now? Not the best idea if i review my own changes :slightly_smiling_face: 
Overall i think this contracts is largely ready. No integration tests done but deployment works fine. But as we have said previously it might be better to hold off on integration Testing until we decided on a structure.

Could add a score interface now but just keep it in the Rebalancing directory for now, or do you want to wait with adding interfaces for contracts until later?

nightowl121:
Sure @AntonAndell I will take a look from here.

core-contracts/Rebalancing/src/main/java/network/balanced/score/core/rebalancing/Rebalancing.java
```diff
@@ -0,0 +1,191 @@
...
+        return sicx.get();
+    }
+
+    @External(readonly = true)
+    public Address getDex() {
+        return dex.get();
+    }
+
+    private BigInteger calculateTokensToSell(BigInteger price, BigInteger baseSupply, BigInteger quoteSupply) {
+        return price.multiply(baseSupply).multiply(quoteSupply).divide(EXA).sqrt().subtract(baseSupply);
```
nightowl121, line 121: This will produce error whenever the price is negative. Can't calculate square root of negative number.

AntonAndell: Yeah saw you changes cleans up nice, it is good the check that the price is not negative since this is a helper function but can it ever be negative?

nightowl121: Oops! got confused with the price difference. Yeah the oracle price can't get negative. I though the price is coming from the difference.

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
```diff
@@ -20,39 +20,47 @@
...
     private Score rebalancingScore;
-    private Score dexMock;
-    private Score loansMock;
-    private Score bnUSDScore;
-    private Score sICXScore;
+    private static final BigInteger FEE = BigInteger.valueOf(997);
+    private static final BigInteger THOUSAND = BigInteger.valueOf(1000);
 
-    private LoansMock loansSpy;
+    private final MockedStatic<Context> contextMock = Mockito.mockStatic(Context.class, Mockito.CALLS_REAL_METHODS);
```
AntonAndell, line 63: Nice! Much better than using mock contracts for unit tests.

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
```diff
@@ -62,395 +70,243 @@ private void expectErrorMessage(Executable _contractCall, String _expectedErrorM
...
+    void setAndGetLoans() {
+        testContractSettersAndGetters("setLoans", loansScore, "getLoans");
     }
 
-    @Test
-    void setLoans_notAdmin() {
-        //Arrange
-        Account sender = sm.createAccount();
-        String expectedErrorMessage = "Rebalancing: Sender not admin";
+    private void testContractSettersAndGetters(String setterMethod, Account scoreAddress, String getterMethod) {
```
AntonAndell, line 125: Also quite clever, However it does do multiple tests in one but is probably fine in such insignificant tests

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
```diff
@@ -62,395 +70,243 @@ private void expectErrorMessage(Executable _contractCall, String _expectedErrorM
...
-        dexMock.invoke(adminAccount, "setPoolStatsBase", poolBase);
-        dexMock.invoke(adminAccount, "setPoolStatsQuote", poolQuote);
-
-        // Act
-        rebalancingScore.invoke(sm.createAccount(), "rebalance");
-        
-        //Assert
-        verify(loansSpy).raisePrice(expectedTokensToSell);
+    public static Stream<BigInteger> provideDifferentThresholds() {
+        return Stream.of(BigInteger.TEN.pow(17), BigInteger.TEN.pow(16));
```
AntonAndell, line 162: I do not understand why you are testing these thresholds in the rebalance method. Could you explain your choice there? Think not using parameterizing here would make more sense

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
nightowl121: At least better than repeating the code. May be, the tests can be changed to parameterized tests, passing different values we want to test.

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
nightowl121: These values represent 10% and 1%. There is no specific reason to choose this specific value, can be changed to other better alternatives.

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
AntonAndell: Yeah, get/sets usually get inherently tested during other tests since the are needed to be set anyway, so not something we should spend time on right now perhaps :smile:

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
AntonAndell: Aah okey then mabye thoose tests can be non parameterized since you have 4 different test doing each case anyway. But doesn't really matter i guess

core-contracts/Rebalancing/src/test/java/network/balanced/score/core/rebalancing/RebalancingTest.java
AntonAndell: On further thought is this smart in complex contract environments, won't we get more value, especially in large contracts, to use mock contracts? 

This is however further down the line and focus right now should be to get continuous rewards and bBaln running.

But when we have Interfaces of all contracts we could easily implement mocks for each contract. where mock contracts implement both a mock interface and a specific contract interface. Thus changing method names/parameters gets reflected across test and does not get outdated. 
When we do not have to write new mocks whenever we unit test a contract it should not be as a large overhead as it is now. 
So basically when we have interfaces we can create a mock class where are methods are default Implemented.

`public class SomeContractMock extends MockContract implements SomeContractInterface {...}`  

Where you only need to define the constructor for SomeContractMock.

Then have MockContract implement methods such as `Call(Address, method, Object... params)` so you can for example could initiate calls from the contract and other good methods. Have not thought this through completely but not sure mocking context will be sustainable in all contracts

AntonAndell: Yeah that looks quite handy, gonna investigate contract mocking with scoreClientInterfaces

AntonAndell: Something like this might be a good replacement once we got interfaces for all contracts.

```
public class MockContract {
    public MockContract() {
    }
}
```

```
SomeScoreClientInterface contractMock = Mockito.mock(SomeScoreClientInterface.class);
Score someContract = sm.deploy(admin, MockContract.class);
someContract.setInstance(contractMock);

// Mock methods
when(contractMock.method()).thenReturn("someValue");

//use Score 
someContract.getAddress();
```

Then if we could abstract it down to some class like

```
MockedContract<SomeScoreClientInterface> someContract = MockedContract.deploy<SomeScoreClientInterface>(sm)
// Inherits Score methods and is mockable
when(someContract.method()).thenReturn("someValue");
someContract.getAddress();
```

This also automatically create default implementations for the interface. So things like 
`contextMock.when(() -> Context.call(eq(loansScore.getAddress()), eq("raisePrice"), any(BigInteger.class))).thenReturn(null); `
are done automatically.

AntonAndell:
@nightowl121 really like the addition of score-lib and test-lib! we are starting to get some infrastructure in place. How soon do you think we can get it to master? so other devs can use and contribute to it?

Do you need any help here or on the Staking contracts? Since those contain most of the infrastructure it would ease development on other contracts to get it in asap.

AntonAndell:
@nightowl121 really like the addition of score-lib and test-lib! we are starting to get some infrastructure in place. How soon do you think we can get it to master? so other devs can use and contribute to it?

Do you need any help here or on the Staking contracts? Since those contain most of the infrastructure it would ease development on other contracts to get it in asap.

nightowl121:
Rebalancing is ready to be merged to master. 

I would definitely need help in having robust integration test framework, starting from Staking contract.

AntonAndell:
Sweet then lets merge it soon. I can do a quick review of you recent changes tonight and perhaps merge after that?

I can start looking a bit on the staking contract as well and start to contribute to the infrastructure there as soon as i've done the first draft of the rewards contract. What do you you feel we are missing currently in the staking contracts and what is missing in the integration test framework.

nightowl121:
I would consider integration test framework as done if we could run a single command and run all the integration tests automatically. Currently the integration tests is not complete in that regards. It can be run manually though.

Once we have automation in that, we could apply CI for running integration tests in local gochain node. We could run one local gochain node and run our integration tests against it.

nightowl121:
> Sweet then lets merge it soon. I can do a quick review of you recent changes tonight and perhaps merge after that?
> 
> I can start looking a bit on the staking contract as well and start to contribute to the infrastructure there as soon as i've done the first draft of the rewards contract. What do you you feel we are missing currently in the staking contracts and what is missing in the integration test framework.

Rebalancing is ready for merge from my side, you may take a look. @AntonAndell

AntonAndell:
It looks good, but adding a scoreclient interface might be good. I can add and push, it should be quick to write

nightowl121:
> It looks good, but adding a scoreclient interface might be good. I can add and push, it should be quick to write

Yeah, Please go ahead.

AntonAndell:
Done, check that and if u don't see anything super weird, merge it!

checked that all tests pass and deployment to my local node works

AntonAndell:
Done, check that and if u don't see anything super weird, merge it!

checked that all tests pass and deployment to my local node works

AntonAndell (approved):

AntonAndell:
> Did you try keeping this compileOnly and annotation processor?

hmm no i can try it. Which one do we want annotation or compileOnly? or do you need both?

AntonAndell (approved):

AntonAndell:
Also changed to Impl naming so when creating interfaces the syntax will be a lot more readable

nightowl121:
Moved the interface to score-lib.

nightowl121 (approved):
Approved

PR #63, opened 2022-02-26
Translation of bnUSD token 

ZeChosenName:


token-contracts/bnUSD/src/test/java/network/balanced/score/tokens/TestBNUSD.java
```diff
@@ -0,0 +1,185 @@
...
+import org.junit.jupiter.api.*;
+import score.Address;
+import score.Context;
+
+import java.math.BigInteger;
+import java.util.Stack;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class TestBNUSD extends TestBase {
```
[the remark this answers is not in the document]
AntonAndell: Also seem to be a lot of test missing, and i feel we should test all functionality in IRC2 implementations aswell.

AntonAndell (changes requested):
Take a look at the offical implementations of IRC2 if you haven't already. https://github.com/sink772/javaee-tokens/find/master
Then since this does not apply fully to all functionalities we need to test the token implementations. 

Also some pointers on code style consistency.
use at max only 1 newline to separate code.
newlines after } unless it is another }
```
if (a) {
     if (b) {
        do x;
    }
}

return x;
```
Only use comment when the code is so bad that it need explanations, and before you add it try to think about how the code can be improved instead. 

Keep up the good work :muscle:  I might be a bit extra hard on testing on this on since this contracts if very bad to bugs in, so try and cover as many cases as possible.

Also if you can try deploying it to find issues such as those with usage of non allowed methods.

Best way to do it is with the esitmate step command, it gives the best error messages.
```
./goloop rpc sendtx deploy /path/to/optimized.jar     --uri http://localhost:9082/api/v3     --key_store godWallet.json --key_password gochain
 --nid 3 --step_limit=100000000     --content_type application/java   --estimate  --param x=y z=u
```

token-contracts/bnUSD/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+    compileOnly 'foundation.icon:javaee-api:0.9.1'
+    testImplementation 'foundation.icon:javaee-unittest:0.9.2'
+
+    testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
+    testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine:5.8.1'
+}
+
+test {
+    useJUnitPlatform()
+}
```
AntonAndell, line 21:
```suggestion
test {
    useJUnitPlatform()
    finalizedBy jacocoTestReport 
}

jacocoTestReport {
    dependsOn test
    reports {
        xml.required = false
        csv.required = false
        html.outputLocation = layout.buildDirectory.dir('jacocoHtml')
    }
}
```

Add this and then to project gradle add 
`apply plugin: 'jacoco'`
and you can get coverage reports of your unitTests. You wont need a 100% but you can look through there to check you are covering all the good spots :slightly_smiling_face:

token-contracts/bnUSD/src/main/java/network/balanced/score/tokens/tokens/IRC2Mintable.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param _amount  Number of tokens to be created at the account.
+     * @param _data
+     */
+    @External
+    public void mint(BigInteger _amount, @Optional byte[] _data) {
+        if (_data == null) {
+            String data = "None";
+            _data = data.getBytes();
+        }
+    }
```
Suyog007, line 36: We need to call _mint(function of IRC2) too .
https://github.com/balancednetwork/balanced-contracts/blob/development/token_contracts/bnUSD/tokens/IRC2mintable.py#L21

token-contracts/bnUSD/src/main/java/network/balanced/score/tokens/tokens/IRC2Mintable.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Creates `_amount` number of tokens, and assigns to `_account`.
+     * 		Increases the balance of that account and total supply.
+     * @param _account The account at which token is to be created.
+     * @param _amount Number of tokens to be created at the account.
+     * @param _data
+     */
+    @External
+    public void mintTo(Address _account, BigInteger _amount, @Optional byte[] _data) {
```
Suyog007, line 46: We need to check if the _data we received is null or not just the way we did in above mint function

token-contracts/bnUSD/src/main/java/network/balanced/score/tokens/tokens/IRC2.java
```diff
@@ -0,0 +1,291 @@
...
+    final private static String NAME = "name";
+    final private static String SYMBOL = "symbol";
+    final private static String DECIMALS = "decimals";
+    final private static String TOTAL_SUPPLY = "total_supply";
+    final private static String BALANCES = "balances";
+    final private static String ADMIN = "admin";
+    final private static String ACCOUNTS = "account";
+
+    static String zero = "0";
+    final protected static Address EOA_ZERO = Address.fromString("hx" + zero.repeat(40));
```
Suyog007, line 20: .repeat should be avoided as it will give an issue while converting to jar file. Instead ,we can directly assign hx000.... to a constant.

token-contracts/bnUSD/src/main/java/network/balanced/score/tokens/tokens/TokenStandard.java
```diff
@@ -0,0 +1,19 @@
...
+    public String symbol();
+
+    public BigInteger decimals();
+
+    public BigInteger totalSupply();
+
+    public BigInteger balanceOf(Address _owner);
+
+    public void transfer(Address _to, BigInteger _value, byte[] _data);
+}
```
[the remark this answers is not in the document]
nightowl121: We have implemented our own due to the difference in variable keys name.

nightowl121:
@AntonAndell I have updated as per your suggestions. Please check the updated PR. The contract is ready for deployment.

token-contracts/bnUSD/src/main/java/network/balanced/score/tokens/tokens/TokenStandard.java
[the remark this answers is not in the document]
AntonAndell: Aah, did not notice that

token-contracts/bnUSD/src/main/java/network/balanced/score/tokens/BalancedDollar.java
```diff
@@ -0,0 +1,205 @@
...
+    public VarDB<Address> oracleAddress = Context.newVarDB(ORACLE_ADDRESS, Address.class);
+    public VarDB<String> oracleName = Context.newVarDB(ORACLE_NAME, String.class);
+    public VarDB<BigInteger> priceUpdateTime = Context.newVarDB(PRICE_UPDATE_TIME, BigInteger.class);
+    public VarDB<BigInteger> lastPrice = Context.newVarDB(LAST_PRICE, BigInteger.class);
+    public VarDB<BigInteger> minInterval = Context.newVarDB(MIN_INTERVAL, BigInteger.class);
+
+
+    public BalancedDollar(@Optional Address _governance) {
+        // null to emulate no args supplied
+        super(TOKEN_NAME, SYMBOL_NAME, null, null);
```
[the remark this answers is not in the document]
nightowl121: Have adjusted all the contracts constructor so that the code is not executed twice during the upgrades.

AntonAndell (approved):
Overall structure of token implementation looks great now! nice work. 
From quick testing and glance i can't find any functional issues, so looks good to me :+1:  Think in some places we could improve even more on code structure and testing but just minor things which i think we shouldn't put more effort into now since there is a lot more work needed elsewhere.

score-lib/src/main/java/network/balanced/score/lib/tokens/IRC2Burnable.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class IRC2Burnable extends IRC2Mintable implements IRC2BurnableInterface {
+
+    public IRC2Burnable(String _name, String _symbol, @Optional BigInteger _decimals) {
+        super(_name, _symbol, _decimals);
+    }
+
+    @External
+    public void burn(BigInteger _amount) {
+        burnFrom(Context.getCaller(), _amount);
```
AntonAndell, line 37: Just curious, do you how would it affect the system if a asset is burned by a user?

AntonAndell (approved):

score-lib/src/main/java/network/balanced/score/lib/tokens/IRC2Burnable.java
nightowl121: bnUSD is generated by locking icx and getting a debt. If we allow the user's to burn bnusd, the debt position still remains open but there is no bnusd to payback the loan. Similarly in case of icx, if we burn sicx, the ratio of sicx:icx in the contract changes creating unfavorable situations.

AntonAndell: Yeah okey it gets a bit weird. Just off the bat burning bnusd seems like it would create a stronger backing. But needs to be handled correctly :+1:

PR #68, opened 2022-03-03
fee handler java translation

simusud:
fee handler java translation and unit tests


core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+import score.annotation.External;
+import score.annotation.Optional;
+import scorex.util.ArrayList;
+import scorex.util.HashMap;
+
+import java.math.BigInteger;
+import java.util.List;
+import java.util.Map;
+
+import static network.balanced.score.core.Checks.onlyGovernance;
```
AntonAndell, line 14: import static network.balanced.score.core.Checks.*;

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+import static network.balanced.score.core.Checks.onlyOwner;
+
+
```
AntonAndell, line 17: no need for double newlines i think

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+    public static final BranchDB<Address, DictDB<Address, String>> routes = Context.newBranchDB("routes", String.class);
+    public static final VarDB<Address> governance = Context.newVarDB("governance", Address.class);
+    public static final VarDB<Boolean> enabled = Context.newVarDB("enabled", Boolean.class);
+    public static final ArrayDB<Address> allowed_address = Context.newArrayDB("allowed_address", Address.class);
+    public static final VarDB<BigInteger> next_allowed_addresses_index = Context.newVarDB("_next_allowed_addresses_index", BigInteger.class);
+
+    public FeeHandler(@Optional Address governance) {
+        if (governance != null) {
+            Context.require(governance.isContract(), "FeeHandler: Governance address should be a contract");
+            FeeHandler.governance.set(governance);
```
AntonAndell, line 33: I perfer usage of this, and i think that is what we are using in other translations

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+    public static final VarDB<BigInteger> fee_processing_interval = Context.newVarDB("block_interval", BigInteger.class);
+    public static final VarDB<byte[]> last_txhash = Context.newVarDB("last_txhash", byte[].class);
+    public static final BranchDB<Address, DictDB<Address, String>> routes = Context.newBranchDB("routes", String.class);
+    public static final VarDB<Address> governance = Context.newVarDB("governance", Address.class);
+    public static final VarDB<Boolean> enabled = Context.newVarDB("enabled", Boolean.class);
+    public static final ArrayDB<Address> allowed_address = Context.newArrayDB("allowed_address", Address.class);
+    public static final VarDB<BigInteger> next_allowed_addresses_index = Context.newVarDB("_next_allowed_addresses_index", BigInteger.class);
+
+    public FeeHandler(@Optional Address governance) {
+        if (governance != null) {
```
AntonAndell, line 31: Minor thing about this constructor is that you can override the governance parameter by updating the contract. Dont know if this is intended or not. 
But if not id suggest also checking this.governance to see if it is already set.

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+    @External
+    public void setAcceptedDividendTokens(Address[] _tokens) {
+        onlyGovernance();
+        Context.require(_tokens.length <= 10, "There can be a maximum of 10 accepted dividend tokens.");
+        for (Address address : _tokens) {
+            Context.require(address.isContract(), TAG + " :Address provided is an EOA address. Only contract addresses are allowed.");
+        }
+
+        while (accepted_dividends_tokens.size() != 0) {
+            accepted_dividends_tokens.pop();
```
AntonAndell, line 63: pop returns the object popped, i think there is a remove Last function as well which does not return the object after each remove.

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    @External(readonly = true)
+    public Map<String, Object> getRoute(Address _fromToken, Address _toToken) {
+        String path = routes.at(_fromToken).getOrDefault(_toToken, "");
+        if (path.equals("")) {
+            return Map.of();
+        }
+        JsonArray pathJson = Json.parse(path).asArray();
+        String[] pa = new String[pathJson.size()];
```
AntonAndell, line 104: pa is for pathArray? pathArray could be a better name. Might be even better ones, like route, orderdPath, orderdPathArray. You probably have better knowledge for choosing the most correct one or something completely different

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+
+    @External(readonly = true)
+    public BigInteger getFeeProcessingInterval() {
+        return fee_processing_interval.get();
+    }
+
+    @External
+    public void tokenFallback(Address _from, BigInteger _value, byte[] _data) {
+        Address sender = Context.getCaller();
+        if (last_txhash.getOrDefault(new byte[0]) == (Context.getTransactionHash())) {
```
AntonAndell, line 127:
```
        if (last_txhash.getOrDefault(new byte[0]) == (Context.getTransactionHash())) {
            return;
        } else if (!timeForFeeProcessing(sender)) {
            return;
        } 

        last_txhash.set(Context.getTransactionHash());
```

This structure would work as well right?

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    @External(readonly = true)
+    public List<Address> get_allowed_address(int offset) {
+        int start = offset;
+        int end = Math.min(allowed_address.size(), offset + 20) - 1;
+        List<Address> address_list = new ArrayList<>();
+        for (int i = start; i < end + 1; i++) {
+            address_list.add(allowed_address.get(i));
+        }
```
AntonAndell, line 157: space after } is nice :slightly_smiling_face:

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+            } else {
+                if (loop_flag && (starting_index.equals(current_index))) {
+                    Context.revert("No fees on the contract.");
+                }
+                current_index = current_index.add(BigInteger.ONE);
+
+                if (!loop_flag) {
+                    loop_flag = true;
+                }
+            }
```
AntonAndell, line 188:
```suggestion
            if (balance.compareTo(BigInteger.ZERO) > 0) {
                break;
            } else  if (loop_flag && (starting_index.equals(current_index))) {
                  Context.revert("No fees on the contract.");
             }
              
              current_index = current_index.add(BigInteger.ONE);
              if (!loop_flag) {
                  loop_flag = true;
              }
```

AntonAndell: else can be removed here also, since both the ifs break execution so we can remove nesting depth by 1.
Also don't use the apply suggestion feature did not do the code properly

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+
+        }
```
AntonAndell, line 190: newline

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+        next_allowed_addresses_index.set(current_index.add(BigInteger.ONE));
+        JsonArray path;
+        try {
+            String route = routes.at(address).getOrDefault(getContractAddress("baln"), "");
+            path = Json.parse(route).asArray();
+        } catch (Exception e) {
```
AntonAndell, line 196: What are we catching?  Could you explain the usage?
In general we want to avoid caching general Expections since we do not get any information on what goes wrong if something unexpectedly happens in that clause. 

Uncatched exceptions still reverts the transaction so all does not need to be catched.

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+        try {
+            if (path.size() > 0) {
+                transferToken(address, getContractAddress("router"), balance, createDataFieldRouter(
+                        getContractAddress("dividends"), path));
+            } else {
+                transferToken(address, getContractAddress("dex"), balance, createDataFieldDex(
+                        getContractAddress("baln"), getContractAddress("dividends")));
+            }
+        } catch (Exception e) {
+            Context.revert("No fees on the contract " + address + " failed" + e.getMessage());
```
AntonAndell, line 209: Same here for the try catch, id rather we catch the specific exception that tells us we do not have any fees, and even better use the Context.require to assert the values needed if possible

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+
+    public Address getContractAddress(String _contract) {
+        return (Address) Context.call(governance.getOrDefault(Checks.defaultAddress), "getContractAddress", _contract);
+    }
+
+    public boolean timeForFeeProcessing(Address _token) {
+        if (enabled.getOrDefault(Boolean.FALSE).equals(false)) {
+            return Boolean.FALSE;
+        }
+        BigInteger blockHeight = BigInteger.valueOf(Context.getBlockHeight());
```
AntonAndell, line 237:
```suggestion
        }
        
        BigInteger blockHeight = BigInteger.valueOf(Context.getBlockHeight());
```

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+        BigInteger last_conversion = last_fee_processing_block.getOrDefault(_token, BigInteger.ZERO);
+        BigInteger target_block = last_conversion.add(fee_processing_interval.getOrDefault(BigInteger.ZERO));
+        if (last_conversion.signum() > 0) {
+            return Boolean.TRUE;
+        } else if (blockHeight.compareTo(target_block) < 0) {
+            return Boolean.FALSE;
+        } else {
+            return Boolean.TRUE;
```
AntonAndell, line 245: Hmm unsure if this needed, i have not had any problems returning primitives. (true, false). Do you have a difference in behavior using primitives instead?

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
```diff
@@ -0,0 +1,256 @@
...
+            } else {
+                transferToken(address, getContractAddress("dex"), balance, createDataFieldDex(
+                        getContractAddress("baln"), getContractAddress("dividends")));
+            }
+        } catch (Exception e) {
+            Context.revert("No fees on the contract " + address + " failed" + e.getMessage());
+        }
+    }
+
+    public byte[] createDataFieldRouter(Address _receiver, JsonArray _path) {
```
AntonAndell, line 213: All of these non external functions should probably be made private. Does not really matter but could allow for unrepresentative behavior during testing.

core-contracts/feehandler/src/main/java/network/balanced/score/core/Checks.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public static void onlyGovernance() {
+        Address sender = Context.getCaller();
+        Address governance = FeeHandler.governance.getOrDefault(defaultAddress);
+        Context.require(!governance.equals(defaultAddress), FeeHandler.TAG + ": Governance address not set");
+        Context.require(sender.equals(governance), FeeHandler.TAG + ": Sender not governance contract");
+    }
+
+}
```
AntonAndell, line 23:
```suggestion
    }
}
```

AntonAndell (changes requested):
Initial review looks good, some mostly minor changes :muscle: . 
Also make sure the rebase on main and use the dependency structure in place there.

Still need to look at the testing, will do that a little later.

AntonAndell:
@simusud If want some help with evaluating the test you can add jacoo test coverage, so you can see what code you cover int the unit tests. you can see it in the loansContract rewrite or on the bnusd pull request i showed how to add it.

simusud:
> Initial review looks good, some mostly minor changes 💪 . Also make sure the rebase on main and use the dependency structure in place there.
> 
> Still need to look at the testing, will do that a little later.
> 
> Edit: I do not see a gradle file commited for you contract, did you forget to add it? Will pause review until that is resolved so i can run the code.

sorry about that , i have just uploaded the gradle file and rebased the branched to main.

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
simusud: it was to catch the case when route is empty. But yeah, i have optimized it removing try and catch

core-contracts/feehandler/src/main/java/network/balanced/score/core/FeeHandler.java
simusud: Same here, will remove try and catch as it looks unnecessary to catch exception.

simusud:
Made the requested changes .  Please have a look once .

core-contracts/feehandler/src/test/java/network/balanced/score/core/feehandler/FeeHandlerTest.java
```diff
@@ -0,0 +1,151 @@
...
+    void getEmptyRoute(){
+        feehandler.invoke(governance.getAccount(), "setRoute", sicxScore.getAddress(), baln.getAddress(), new Address[] {bnusd.getAddress()});
+        assertEquals(Map.of(), feehandler.call("getRoute", bnusd.getAddress(), baln.getAddress()));
+    }
+
+    @Test
+    void deleteRoute(){
+        feehandler.invoke(governance.getAccount(), "setRoute", sicxScore.getAddress(), baln.getAddress(), new Address[] {bnusd.getAddress()});
+        feehandler.call("getRoute", sicxScore.getAddress(), baln.getAddress());
+        feehandler.invoke(governance.getAccount(), "deleteRoute",sicxScore.getAddress(), baln.getAddress());
```
AntonAndell, line 91: Needs to assert that it is deleted

core-contracts/feehandler/src/test/java/network/balanced/score/core/feehandler/FeeHandlerTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertEquals(BigInteger.TEN, feehandler.call("getFeeProcessingInterval"));
+    }
+
+    @Test
+    void tokenFallback(){
+        setAcceptedDividendTokens_NoPreviousTokens();
+        setFeeProcessingInterval();
+        feehandler.invoke(governance.getAccount(), "enable");
+        feehandler.invoke(sicxScore.getAccount(), "tokenFallback", baln.getAddress(), BigInteger.TEN.pow(2), new byte[0]);
+    }
```
AntonAndell, line 106: Needs assert, to show what it is testing

core-contracts/feehandler/src/test/java/network/balanced/score/core/feehandler/FeeHandlerTest.java
```diff
@@ -0,0 +1,151 @@
...
+        feehandler.invoke(owner, "route_contract_balances");
+    }
+
+    @Test
+    void route_contract_balances_with_empty_path(){
+        bnusd.invoke(owner, "mintTo", feehandler.getAddress(), MINT_AMOUNT);
+        feehandler.invoke(owner, "add_allowed_address", bnusd.getAddress());
+        setGetRoute();
+        feehandler.invoke(owner, "route_contract_balances");
+    }
```
AntonAndell, line 139: Assert that the correct contracts are called, id use mockito spy to achieve this.

core-contracts/feehandler/src/test/java/network/balanced/score/core/feehandler/FeeHandlerTest.java
```diff
@@ -0,0 +1,151 @@
...
+    @Test
+    void get_allowed_address_NegativeOffset(){
+        feehandler.invoke(owner, "add_allowed_address", sicxScore.getAddress());
+        try{
+            feehandler.call("get_allowed_address", -1);
+        }catch (AssertionError e){
+            assertEquals("Reverted(0): Negative value not allowed.", e.getMessage());
+        }
+    }
+
```
AntonAndell, line 125: Add a test for route_contract_balances to check for infinite loops, a list larget than 1 with all balance 0 i think

core-contracts/feehandler/src/test/java/network/balanced/score/core/feehandler/FeeHandlerTest.java
```diff
@@ -0,0 +1,151 @@
...
+        expc.add(sicxScore.getAddress());
+        expc.add(baln.getAddress());
+        assertEquals(expc, feehandler.call("getAcceptedDividendTokens"));
+    }
+
+    @Test
+    void setGetRoute() {
+        feehandler.invoke(governance.getAccount(), "setRoute", sicxScore.getAddress(), baln.getAddress(), new Address[] {bnusd.getAddress()});
+        Map<String, Object> res = (Map<String, Object>) feehandler.call("getRoute", sicxScore.getAddress(), baln.getAddress());
+        assertEquals(baln.getAddress(), res.get("toToken"));
```
AntonAndell, line 78: Assert the other 2 parameters aswell

core-contracts/feehandler/src/test/java/network/balanced/score/core/feehandler/FeeHandlerTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+import com.iconloop.score.test.Account;
+import com.iconloop.score.test.Score;
+import com.iconloop.score.test.ServiceManager;
+import com.iconloop.score.test.TestBase;
+import network.balanced.score.core.feehandler.utils.*;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import score.Address;
+import scorex.util.ArrayList;
```
AntonAndell, line 11: U can use java.util.arrayList here

AntonAndell (changes requested):
Contract look almost finished, just to comments.

Testing can still be improved i think.
Missing a few test cases and could improve structure a bit. Please just tag me in comments here if you have any questions/don't agree. I usually respond quite quickly.
Try to keep a consistent structure, usually

Setup all you need for the test, variables/contract state

Do something, call methods/transfer tokens

Verify everything relevant, Assert that everything is as it is supposed to be. The correct values are set or the correct methods are called

AntonAndell:
Look at how nightowl use static mock to replace the need to mock normal contracts.
https://github.com/balancednetwork/balanced-java-contracts/pull/54#discussion_r837346066

You can look there to see how you can probably remove the need for a governance mock

AntonAndell:
Developing now is like constantly developing against a moving target, sorry about that. But if you can please rebase on master, and add a interface to you contract in score-lib, which you then implement in your contract. You can check how it is done for example rebalancing.

nightowl121:
@AntonAndell feehandler is ready for merge

AntonAndell (approved):
Looks good, but admin is used instead of governance in many places. Is that an active choice and if it is are we sure governance is admin on mainnet?

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    @External(readonly = true)
+    public boolean isEnabled() {
+        return enabled.getOrDefault(false);
+    }
+
+    @External
+    public void setAcceptedDividendTokens(Address[] _tokens) {
+        only(admin);
```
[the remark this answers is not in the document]
nightowl121: Yeah, only(admin) is an intended choice to bring uniformity with other contracts. Currently admin is not set in mainnet.  It can be set either during upgrade or later through governance contract. All the methods having only(admin) are configuration methods which are already being set in mainnet, so they don't have immediate use also.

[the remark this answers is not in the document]
AntonAndell: Sounds good

nightowl121 (approved):
Approved

nightowl121 (approved):
Approved

PR #70, opened 2022-03-08
Use buildSrc to manage the dependencies

nightowl121:
* buildSrc will also be used to manage other complex tasks in Java and also manage constants.
* Custom plugins and tasks can be created in buildSrc folder
* Changing content in buildSrc invalidates the cache of build, this can be improved later by using composite build


simusud (commented):
Looks good

simusud (approved):

PR #71, opened 2022-03-09
[Do not merge] Only for discussion

nightowl121:


core-contracts/staking/src/main/java/network/balanced/score/core/db/IdFactory.java
```diff
@@ -0,0 +1,20 @@
...
+    private static final String NAME = "_ID_FACTORY";
+    private final VarDB<BigInteger> uid;
+    public IdFactory(String key) {
+        String name = key + NAME;
+        this.uid = Context.newVarDB(name + "_uid", BigInteger.class);
+    }
+
+    public BigInteger getUid(){
+        uid.set(uid.getOrDefault(BigInteger.ZERO).add(BigInteger.ONE));
+        return uid.getOrDefault(BigInteger.ZERO);
```
nightowl121, line 18: Reduce get call by introducing local variable.

core-contracts/staking/src/main/java/network/balanced/score/core/db/NodeDB.java
```diff
@@ -0,0 +1,100 @@
...
+    private static final BigInteger UNINITIALIZED = BigInteger.ZERO;
+    private static final BigInteger INITIALIZED = BigInteger.ONE;
+
+    private final VarDB<BigInteger> value;
+    private final VarDB<BigInteger> init;
+    private final VarDB<BigInteger> blockHeight;
+    private final VarDB<BigInteger> next;
+    private final VarDB<BigInteger> prev;
+    private final VarDB<Address> key;
+    private final VarDB<Address> senderAddress;
```
nightowl121, line 21: Preserve order of variables from previous python contract

core-contracts/staking/src/main/java/network/balanced/score/core/db/NodeDB.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public NodeDB(String key) {
+        String name = key + NAME;
+        this.value = Context.newVarDB(name + "_value", BigInteger.class);
+        this.blockHeight = Context.newVarDB(name + "_block_height", BigInteger.class);
+        this.init = Context.newVarDB(name + "_init", BigInteger.class);
+        this.next = Context.newVarDB(name + "_next", BigInteger.class);
+        this.prev = Context.newVarDB(name + "_prev", BigInteger.class);
+        this.key = Context.newVarDB(name + "_key", Address.class);
+        this.senderAddress = Context.newVarDB(name + "_sender_address", Address.class);
```
nightowl121, line 31: Key mismatch from python contract.

core-contracts/staking/src/main/java/network/balanced/score/core/db/NodeDB.java
```diff
@@ -0,0 +1,100 @@
...
+        key.set(null);
+        blockHeight.set(null);
+        senderAddress.set(null);
+        prev.set(null);
+        next.set(null);
+        init.set(null);
+    }
+
+    public boolean exists(){
+        return ( ! init.getOrDefault(BigInteger.ZERO).equals(NodeDB.UNINITIALIZED));
```
nightowl121, line 45:
```java
return init.get() != null
```

Use this and remove the unitialized variable

core-contracts/staking/src/main/java/network/balanced/score/core/db/NodeDB.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public void setBlockHeight(BigInteger blockHeight){
+        init.set(NodeDB.INITIALIZED);
+        this.blockHeight.set(blockHeight);
+    }
+
+    public void setSenderAddress(Address senderAddress){
+        init.set(NodeDB.INITIALIZED);
+        this.senderAddress.set(senderAddress);
+    }
```
nightowl121, line 82: Since all values are set once, make a single method to set all values.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    public void delete() throws Exception {
+        clear();
+        headId.set(null);
+        tailId.set(null);
+        length.set(null);
+    }
+
+    public BigInteger __len__(){
```
nightowl121, line 35: Rename this method to size()

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,208 @@
...
+        }
+        return getNode(tailId);
+    }
+
+    public Object[] createNode(Address key, BigInteger value,
+                               BigInteger blockHeight, Address senderAddres,
+                               BigInteger nodeId) throws Exception {
+        if (nodeId == null) {
+            nodeId = new IdFactory(name + "_nodedb").getUid();
+        }
```
nightowl121, line 96: This block can be removed, and hence id factory can also be removed. There seems to be no usage of IdFactory.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,208 @@
...
+        headId.set(null);
+        tailId.set(null);
+        length.set(null);
+    }
+
+    public BigInteger __len__(){
+        return length.getOrDefault(BigInteger.ZERO);
+    }
+
+    public NodeDB node(BigInteger nodeId){
```
nightowl121, line 39: Rename the method to createNodeInstance

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,208 @@
...
+    public NodeDB getNode(BigInteger nodeId) throws Exception {
+        NodeDB node = node(nodeId);
+        if (! node.exists()){
+            LinkedNodeNotFound(name, nodeId);
+        }
+        return node;
+    }
+
+    public NodeDB getTailNode() throws Exception {
+        BigInteger tailId = this.tailId.getOrDefault(BigInteger.ZERO);
```
nightowl121, line 84: Use get() only. 
Using getOrDefault is mistake here since we are checking for null value.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,208 @@
...
+            NodeDB curprev = getNode(curPrevId);
+            curnext.setPrev(curPrevId);
+            curprev.setNext(curNextId);
+            cur.delete();
+            length.set(length.getOrDefault(BigInteger.ZERO).subtract(BigInteger.ONE));
+        }
+    }
+
+    public void clear() throws Exception {
+        BigInteger curId = headId.getOrDefault(BigInteger.ZERO);
```
nightowl121, line 157: Use get() only.
Using getOrDefault is mistake here since we are checking for null value.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+        while (!curId.equals(tailId)) {
+            curId = node.getNext();
+            node.delete();
+            node = getNode(curId);
+        }
+        node.delete();
+
+        this.tailId.set(null);
+        headId.set(null);
+        length.set(BigInteger.ZERO);
```
nightowl121, line 177: Set length as null value

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+
+    public NodeDB createNodeInstance(BigInteger nodeId) {
+        return new NodeDB(nodeId.toString() + name);
+    }
+
+    public void updateNode(Address key, BigInteger value, BigInteger blockHeight, Address senderAddress,
+                           BigInteger nodeId) {
+        NodeDB node = createNodeInstance(nodeId);
+        if (node.exists()) {
+            node.setter(key, value, blockHeight, senderAddress);
```
nightowl121, line 63: rename it to setValues()

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+
+        } else {
+            Context.revert("There is no node of the provided node id.");
+        }
+    }
+
+    public BigInteger append(Address key, BigInteger value, BigInteger blockHeight
+            , Address senderAddress, BigInteger nodeId) {
+        NodeDB cur = createNode(key, value, blockHeight, senderAddress, nodeId);
+        if (length.getOrDefault(BigInteger.ZERO).equals(BigInteger.ZERO)) {
```
nightowl121, line 73: check for null value rather than zero.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+            node.setter(key, value, blockHeight, senderAddress);
+
+        } else {
+            Context.revert("There is no node of the provided node id.");
+        }
+    }
+
+    public BigInteger append(Address key, BigInteger value, BigInteger blockHeight
+            , Address senderAddress, BigInteger nodeId) {
+        NodeDB cur = createNode(key, value, blockHeight, senderAddress, nodeId);
```
nightowl121, line 72: rename it to `nodeToAppend`

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+            clear();
+        } else {
+            NodeDB oldHead = getNode(headId.getOrDefault(BigInteger.ZERO));
+            BigInteger newHead = oldHead.getNext();
+            headId.set(newHead);
+            getNode(newHead).setPrev(BigInteger.ZERO);
+            oldHead.delete();
+            length.set(length.getOrDefault(BigInteger.ZERO).subtract(BigInteger.ONE));
+        }
+
```
nightowl121, line 127: Handle 0 length case. Return in case of zero.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    public void removeTail() {
```
nightowl121, line 130: Can be changed to get. null value is handled in getNode

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+        if (length.getOrDefault(BigInteger.ZERO).equals(BigInteger.ONE)) {
+            clear();
+        } else {
```
nightowl121, line 133: Set it to null

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+    }
+
+    public void remove(BigInteger curId) {
+        if (curId.equals(headId.getOrDefault(BigInteger.ZERO))) {
+            removeHead();
+        } else if (curId.equals(tailId.getOrDefault(BigInteger.ZERO))) {
+            removeTail();
+        } else {
+            NodeDB cur = getNode(curId);
```
nightowl121, line 149: Set null, it will gain steps

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+        } else {
+            NodeDB oldTail = getNode(tailId.get());
+            BigInteger newTail = oldTail.getPrev();
+            tailId.set(newTail);
+            getNode(newTail).setNext(BigInteger.ZERO);
+            oldTail.delete();
+            length.set(length.getOrDefault(BigInteger.ZERO).subtract(BigInteger.ONE));
+        }
+    }
+
```
nightowl121, line 142: In case of zero size, return.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+            removeTail();
+        } else {
+            NodeDB cur = getNode(curId);
+            BigInteger curNextId = cur.getNext();
+            NodeDB curnext = getNode(curNextId);
+            BigInteger curPrevId = cur.getPrev();
+            NodeDB curprev = getNode(curPrevId);
+            curnext.setPrev(curPrevId);
+            curprev.setNext(curNextId);
+            cur.delete();
```
nightowl121, line 156: Introduce BigInteger.ZERO as default value constant.

core-contracts/staking/src/main/java/network/balanced/score/core/db/LinkedListDB.java
```diff
@@ -0,0 +1,210 @@
...
+            length.set(length.getOrDefault(BigInteger.ZERO).subtract(BigInteger.ONE));
+        }
+    }
+
+    public void clear() {
```
nightowl121, line 161: Rename to `nodeToRemove`

core-contracts/staking/src/main/java/network/balanced/score/core/utils/Constant.java
```diff
@@ -0,0 +1,37 @@
...
+    public static final BigInteger HUNDRED = BigInteger.valueOf(100L);
+    public static final BigInteger TOP_PREP_COUNT = BigInteger.valueOf(100L);
+    public static final BigInteger DEFAULT_UNSTAKE_BATCH_LIMIT = BigInteger.valueOf(200L);
+    public static final BigInteger MAX_ITERATION_LOOP = BigInteger.valueOf(100L);
+    public static final BigInteger DEFAULT_DECIMAL_VALUE = BigInteger.valueOf(18L);
+    long calc = (long) Math.pow(2, 256);
+    long newCalc = calc - 1L;
+    public final BigInteger DEFAULT_CAP_VALUE = BigInteger.valueOf(newCalc);
+
+//    DEFAULT_CAP_VALUE = 2 ** 256 -1
```
nightowl121, line 36: Remove this unused constant. If you want to define max value. Refer here:
https://github.com/icon-project/goloop/commit/19cdc39c4e25af2085fb68069f26d991ad80f362#diff-6f6b07de414a79ba16b637420d8f1365444a2e20d5abd922c8ecd79c576e0402R44

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+
+    @EventLog(indexed = 2)
+    public void UnstakeAmountTransfer(Address receiver, BigInteger amount) {
+    }
+
+    @EventLog(indexed = 2)
+    public void IscoreClaimed(BigInteger block_height, BigInteger rewards) {
+    }
+
+    private static final String SICX_SUPPLY = "sICX_supply";
```
nightowl121, line 85: Move these string constants to constant file.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    private static final String PREP_LIST = "_prep_list";
+    private static final String ADDRESS_DELEGATIONS = "_address_delegations";
+    private static final String PREP_DELEGATIONS = "_prep_delegations";
+    private static final String TOTAL_UNSTAKE_AMOUNT = "_total_unstake_amount";
+    private static final String UNSTAKE_BATCH_LIMIT = "_unstake_batch_limit";
+    private static final String STAKING_ON = "staking_on";
+    private static final String ICX_PAYABLE = "icx_payable";
+    private static final String ICX_TO_CLAIM = "icx_to_claim";
+
+    private final VarDB<BigInteger> sicxSupply = Context.newVarDB(SICX_SUPPLY, BigInteger.class);
```
nightowl121, line 105: Move these variables above constructor.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    private final VarDB<BigInteger> totalUnstakeAmount = Context.newVarDB(TOTAL_UNSTAKE_AMOUNT, BigInteger.class);
+    private final ArrayDB<Address> topPreps = Context.newArrayDB(TOP_PREPS, Address.class);
+    private final ArrayDB<Address> prepList = Context.newArrayDB(PREP_LIST, Address.class);
+    private final VarDB<BigInteger> icxToClaim = Context.newVarDB(ICX_TO_CLAIM, BigInteger.class);
+    private final DictDB<String, String> addressDelegations = Context.newDictDB(ADDRESS_DELEGATIONS, String.class);
+    private final DictDB<Address, BigInteger> icxPayable = Context.newDictDB(ICX_PAYABLE, BigInteger.class);
+    private final DictDB<String, BigInteger> prepDelegations = Context.newDictDB(PREP_DELEGATIONS, BigInteger.class);
+    private final VarDB<BigInteger> unstakeBatchLimit = Context.newVarDB(UNSTAKE_BATCH_LIMIT, BigInteger.class);
+    public static final VarDB<Boolean> stakingOn = Context.newVarDB(STAKING_ON, Boolean.class);
+    private final LinkedListDB linkedListDb = new LinkedListDB("unstake_dict");
```
nightowl121, line 123: Rename it to unstakeRequestList

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+import scorex.util.ArrayList;
+import scorex.util.HashMap;
+import scorex.util.StringTokenizer;
+
+
+import static network.balanced.score.core.utils.Checks.*;
+
+public class Staking {
+
+    public Staking() {
```
nightowl121, line 47: Introduce a parameter to avoid calling the below code while updating the contract.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        Map<String, Object> termDetails =
+                (Map<String, Object>) Context.call(Address.fromString(Constant.SYSTEM_SCORE_ADDRESS), "getIISSInfo");
```
nightowl121, line 49: Define system score address as address constant.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+import scorex.util.HashMap;
+import scorex.util.StringTokenizer;
+
+
+import static network.balanced.score.core.utils.Checks.*;
+
+public class Staking {
+
+    public Staking() {
+        Map<String, Object> termDetails =
```
nightowl121, line 48: Use score interface to do interscore calls.

core-contracts/staking/src/main/java/network/balanced/score/core/utils/Constant.java
```diff
@@ -0,0 +1,37 @@
...
+package network.balanced.score.core.utils;
+
+import java.math.BigInteger;
+import java.lang.Math;
+
+
+public class Constant {
+    public static final String TAG = "Staked ICX Manager";
+    public static final String SYSTEM_SCORE_ADDRESS = "cx0000000000000000000000000000000000000000";
+    public static final BigInteger DENOMINATOR = BigInteger.valueOf(1000000000000000000L);
```
nightowl121, line 26: Use power function to define constant.

core-contracts/staking/src/main/java/network/balanced/score/core/utils/Constant.java
```diff
@@ -0,0 +1,37 @@
...
+    public static final BigInteger HUNDRED = BigInteger.valueOf(100L);
+    public static final BigInteger TOP_PREP_COUNT = BigInteger.valueOf(100L);
```
nightowl121, line 28: Use the HUNDRED defined above.

core-contracts/staking/src/main/java/network/balanced/score/core/utils/Constant.java
```diff
@@ -0,0 +1,37 @@
...
+    public static final BigInteger DEFAULT_UNSTAKE_BATCH_LIMIT = BigInteger.valueOf(200L);
+    public static final BigInteger MAX_ITERATION_LOOP = BigInteger.valueOf(100L);
```
nightowl121, line 30: Change its data type to Integer

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    }
+
+    @External(readonly = true)
+    public BigInteger getBlockHeightWeek() {
+        return blockHeightWeek.getOrDefault(BigInteger.ZERO);
+    }
+
+    @External(readonly = true)
+    public BigInteger getTodayRate() {
+        return rate.getOrDefault(BigInteger.ZERO);
```
nightowl121, line 144: Use get only

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            delegationIcx.put(name, votesIcx);
+        }
+        return delegationIcx;
+    }
+
+    public BigInteger getRate() {
+        BigInteger totalStake = this.totalStake.getOrDefault(BigInteger.ZERO);
+        BigInteger totalSupply = (BigInteger) Context.call(sicxAddress.get(), "totalSupply");
+        BigInteger rate;
+        if (totalStake.compareTo(BigInteger.ZERO) == 0) {
```
nightowl121, line 374: Use `.equals` for comparispon.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+
+    @External
+    public void setUnstakeBatchLimit(BigInteger _limit) {
+        onlyOwner();
+        unstakeBatchLimit.set(_limit);
+    }
+
+    @External(readonly = true)
+    public BigInteger getUnstakeBatchLimit() {
+        return unstakeBatchLimit.getOrDefault(BigInteger.ZERO);
```
nightowl121, line 166: Default value is `DEFAULT_UNSTAKE_BATCH_LIMIT`

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            }
+            topPreps.add(prepAddress);
+        }
+    }
+
+    @External(readonly = true)
+    public Map<String, BigInteger> getAddressDelegations(Address _address) {
+        Map<String, BigInteger> delegationIcx = new HashMap<>();
+        Map<String, BigInteger> delegationPercent = new HashMap<>();
+        delegationPercent = delegationInPer(_address);
```
nightowl121, line 358: define and assign in same line.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        BigInteger balance = (BigInteger) Context.call(sicxAddress.get(), "balanceOf", _address);
+        BigInteger totalIcxHold = (balance.multiply(rate.getOrDefault(BigInteger.ZERO))).divide(Constant.DENOMINATOR);
```
nightowl121, line 360: Use getRate method.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        for (String name : delegationPercent.keySet()) {
```
nightowl121, line 361: Rename `name` to `prepAddress`

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
Suyog007: It should use getTodayRate method.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            StringTokenizer st = new StringTokenizer(delegationString, "\\.");
+            Map<String, BigInteger> delegationPercent = new HashMap<>();
+            List<String> splittedList = new ArrayList<>();
+            splittedList = splitResult(st);
+            for (String item : splittedList) {
+                st = new StringTokenizer(item, ":");
+                splittedList = splitResult(st);
+                if (delegationPercent.get(splittedList.get(0)) == null) {
+                    delegationPercent.put(splittedList.get(0), new BigInteger(splittedList.get(1)));
+                } else {
```
nightowl121, line 667: Remove double looping, avoid creation of array list.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+                    if ((splittedList.get(1) != "0")) {
+                        BigInteger value = delegationPercent.get(splittedList.get(0));
+                        delegationPercent.put(splittedList.get(0), new BigInteger(splittedList.get(1)).add(value));
+                    } else {
+                        delegationPercent.put(splittedList.get(0), BigInteger.ZERO);
```
nightowl121, line 672: Only this condition is sufficient provided address is not written twice in the delegation string.

nightowl121: Use String tokenizer iterator

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        }
+        List<Map<String, Object>> result = (List<Map<String, Object>>) stakeInNetwork.get("unstakes");
+        Map<String, Object> recentUnstakeInfo = result.get(result.size() - 1);
+        linkedListDb.append(to, amountToUnstake,
+                (BigInteger) recentUnstakeInfo.get("unstakeBlockHeight"),
+                addressToSend,
+                linkedListDb.tailId.getOrDefault(BigInteger.ZERO).add(BigInteger.ONE));
+        sicxSupply.set(sicxSupply.getOrDefault(BigInteger.ZERO).subtract(value));
+        UnstakeRequest(addressToSend, amountToUnstake);
+    }
```
nightowl121, line 753: This method is not scalable, its recommended to use pagination.

nightowl121: Avoid creation of new array list.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    @SuppressWarnings("unchecked")
+    public void setTopPreps() {
+        Map<String, Object> prepDict =
+                (Map<String, Object>) Context.call(Address.fromString(Constant.SYSTEM_SCORE_ADDRESS), "getPReps", 1,
+                        Constant.TOP_PREP_COUNT);
+        List<Map<String, Object>> prepDetails = (List<Map<String, Object>>) prepDict.get("preps");
+        List<Address> addresses = getPrepList();
+        for (Map<String, Object> preps : prepDetails) {
+            Address prepAddress = (Address) preps.get("address");
+            if (!contains(prepAddress, addresses)) {
```
nightowl121, line 346: Use contains of `java/util/List`

nightowl121: Use EnumerableSetDB to store prep list later.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            }
+        }
+        return false;
+    }
+
+    public BigInteger delegateVotes(Address to, PrepDelegations[] userDelegations, BigInteger userIcxHold) {
+        BigInteger amountToStake = BigInteger.ZERO;
+        List<Address> similarPrepCheck = new ArrayList<>();
+        StringBuilder addressDelegations = new StringBuilder();
+        List<Address> addresses = getPrepList();
```
nightowl121, line 395: Rename it to allPrepAddresses

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        for (PrepDelegations singlePrep : userDelegations) {
+            Address prepAddress = singlePrep._address;
+            BigInteger votesInPer = singlePrep._votes_in_per;
+            if (!contains(prepAddress, addresses)) {
```
nightowl121, line 399: Aplpies same comments as `setTopPreps`

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+                addresses.add(prepAddress);
+                prepList.add(prepAddress);
+            }
+            if (similarPrepCheck.toString().contains(prepAddress.toString())) {
```
nightowl121, line 403: Use contains of `List` directly instead of converting to string.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+                Context.revert(Constant.TAG + ": You can not delegate same Prep twice in a transaction.Your " +
+                        "delegation preference is" + userDelegations);
+            }
+            if (votesInPer.compareTo(BigInteger.valueOf(1000000000000000L)) < 0) {
```
nightowl121, line 407: Use constant for MINIMUM_DELEGATION_PERCENTAGE

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            if (address.equals(target)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public BigInteger delegateVotes(Address to, PrepDelegations[] userDelegations, BigInteger userIcxHold) {
+        BigInteger amountToStake = BigInteger.ZERO;
+        List<Address> similarPrepCheck = new ArrayList<>();
```
nightowl121, line 393: Use HashSet to check for double prep in delegation preference of user

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            }
+            if (votesInPer.compareTo(BigInteger.valueOf(1000000000000000L)) < 0) {
+                Context.revert(Constant.TAG + ": You should provide delegation percentage more than 0.001. Your " +
+                        "delegation preference is " + userDelegations + ".");
+            }
+            similarPrepCheck.add(prepAddress);
+            amountToStake = amountToStake.add(votesInPer);
+            addressDelegations.append(prepAddress).append(":").append(votesInPer).append(".");
+            setAddressDelegations(to, prepAddress, votesInPer, userIcxHold);
+        }
```
nightowl121, line 415: Instead of returning the totalPercentage, check here if total percentage is equal to 100 or not. Do this check before setting the user address delegation.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        this.addressDelegations.set(to.toString(), addressDelegations.toString());
+        return amountToStake;
+
+    }
+
+    public BigInteger distributeEvenly(BigInteger amountToDistribute, BigInteger isFirstTx, Address to) {
+        BigInteger value = BigInteger.ZERO;
```
nightowl121, line 422: Use length of top preps rather than top prep count as the network can have less than 100 top preps.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            }
+            similarPrepCheck.add(prepAddress);
+            amountToStake = amountToStake.add(votesInPer);
+            addressDelegations.append(prepAddress).append(":").append(votesInPer).append(".");
+            setAddressDelegations(to, prepAddress, votesInPer, userIcxHold);
+        }
+        this.addressDelegations.set(to.toString(), addressDelegations.toString());
+        return amountToStake;
+
+    }
```
nightowl121, line 419: Write two methods for the if else code as they have no common part.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+
+    public BigInteger distributeEvenly(BigInteger amountToDistribute, BigInteger isFirstTx, Address to) {
+        BigInteger value = BigInteger.ZERO;
+        if (isFirstTx.equals(BigInteger.ONE)) {
+            BigInteger evenlyDistribution = amountToDistribute.divide(Constant.TOP_PREP_COUNT);
+            StringBuilder addressDelegation = new StringBuilder();
+            BigInteger balance = (BigInteger) Context.call(sicxAddress.get(), "balanceOf", to);
```
nightowl121, line 426: use getTodayRate instead of accessing from vardb.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        }
+        this.addressDelegations.set(to.toString(), addressDelegations.toString());
+        return amountToStake;
+
+    }
+
+    public BigInteger distributeEvenly(BigInteger amountToDistribute, BigInteger isFirstTx, Address to) {
+        BigInteger value = BigInteger.ZERO;
+        if (isFirstTx.equals(BigInteger.ONE)) {
+            BigInteger evenlyDistribution = amountToDistribute.divide(Constant.TOP_PREP_COUNT);
```
nightowl121, line 424: Introduce a new method `getUserICXBalance`

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            StringBuilder addressDelegation = new StringBuilder();
+            BigInteger balance = (BigInteger) Context.call(sicxAddress.get(), "balanceOf", to);
+            BigInteger userIcxBalance =
+                    (balance.multiply(rate.getOrDefault(BigInteger.ZERO))).divide(Constant.DENOMINATOR);
+            for (int i = 0; i < this.topPreps.size(); i++) {
+                Address prep = this.topPreps.get(i);
+                addressDelegation.append(prep).append(":").append(evenlyDistribution).append(".");
+                setAddressDelegations(to, prep, evenlyDistribution, userIcxBalance);
```
nightowl121, line 432: Get the remainder from total 100 percentage/topprepscount and add it to any one prep.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            }
+            addressDelegations.set(to.toString(), addressDelegation.toString());
+        } else {
+            BigInteger evenlyDistribution =
```
nightowl121, line 436: Keep it in a single line.
Use topPreps length to divide.
Return result and remainder from this method. Handle the remainder in the calling method.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    public void setSicxAddress(Address _address) {
+        onlyOwner();
+        Context.require(_address.isContract(), Constant.TAG + ": Address provided is an EOA address. A contract " +
+                "address is required.");
+        sicxAddress.set(_address);
+    }
+
+    public BigInteger percentToIcx(BigInteger votingPercentage, BigInteger totalAmount) {
+        BigInteger numerator = votingPercentage.multiply(totalAmount);
+        return numerator.divide(Constant.DENOMINATOR.multiply(Constant.HUNDRED));
```
nightowl121, line 224: Use constant HUNDRED_PERCENTAGE

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    }
+
+    public void setAddressDelegations(Address to, Address prep, BigInteger votesInPer, BigInteger totalIcxHold) {
+        int checkVal = totalIcxHold.compareTo(BigInteger.ZERO);
+        if (checkVal != 0) {
+            BigInteger value = percentToIcx(votesInPer, totalIcxHold);
+            setPrepDelegations(prep, value);
+        }
+    }
+
```
nightowl121, line 234: Accept prep parameter as string.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            Address prep = this.prepList.get(i);
+            if (!contains(prep, addresses)) {
+                BigInteger prepDelegations = this.prepDelegations.getOrDefault(prep.toString(), BigInteger.ZERO);
+                toDistribute = toDistribute.add(prepDelegations);
+            }
+        }
+        return distributeEvenly(toDistribute, BigInteger.ZERO, null);
+    }
+
+    public Map<String, BigInteger> removePreviousDelegations(Address to) {
```
nightowl121, line 462: Rename delegationInPer to getDelegationsInPercentage.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        String addressStr = to.toString();
+        Map<String, BigInteger> previousDelegations = delegationInPer(to);
+        BigInteger balance = (BigInteger) Context.call(sicxAddress.get(), "balanceOf", to);
+        BigInteger icxHoldPreviously =
+                (balance.multiply(rate.getOrDefault(BigInteger.ZERO))).divide(Constant.DENOMINATOR);
```
nightowl121, line 467: Set null instead of empty string. It earns step.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        if (!previousDelegations.isEmpty()) {
+            addressDelegations.set(addressStr, "");
+            for (String prep : previousDelegations.keySet()) {
```
nightowl121, line 470: Rename votesPer

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+                BigInteger prepDelegations = this.prepDelegations.getOrDefault(prep, BigInteger.ZERO);
```
nightowl121, line 471: Use percentToIcx

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+                BigInteger votesPer = previousDelegations.get(prep);
+                BigInteger votesIcx =
```
nightowl121, line 473: Use a safety check so that prepDelegations is not negative.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+
+    public void stakeAndDelegate(BigInteger evenlyDistributeValue) {
+        stake(getTotalStake());
+        delegations(evenlyDistributeValue);
+    }
+
+    public BigInteger resetTopPreps() {
+        BigInteger toDistribute = BigInteger.ZERO;
+        List<Address> addresses = getTopPreps();
+        for (int i = 0; i < this.prepList.size(); i++) {
```
nightowl121, line 452: Use contains of List

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            value = evenlyDistribution.divide(Constant.DENOMINATOR);
+
+        }
+        return value;
+    }
+
+    public void stakeAndDelegate(BigInteger evenlyDistributeValue) {
+        stake(getTotalStake());
+        delegations(evenlyDistributeValue);
+    }
```
nightowl121, line 447: Rename it to some name like calculateICXOutOfTopPrep

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+
+    public BigInteger resetTopPreps() {
+        BigInteger toDistribute = BigInteger.ZERO;
+        List<Address> addresses = getTopPreps();
+        for (int i = 0; i < this.prepList.size(); i++) {
+            Address prep = this.prepList.get(i);
+            if (!contains(prep, addresses)) {
+                BigInteger prepDelegations = this.prepDelegations.getOrDefault(prep.toString(), BigInteger.ZERO);
+                toDistribute = toDistribute.add(prepDelegations);
+            }
```
nightowl121, line 457: hanlde the remainder also.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    public void fallback() {
+        stakeICX(Context.getCaller(), null);
+    }
+
+    @External
+    public void tokenFallback(Address _from, BigInteger _value, byte[] _data) {
+        stakingOn();
+        if (!Context.getCaller().equals(sicxAddress.get())) {
+            Context.revert(Constant.TAG + ": The Staking contract only accepts sICX tokens." + Context.getCaller() +
+                    " or " + sicxAddress.get());
```
nightowl121, line 260: I think there is no requirement of a try except block.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            balance = balance.subtract(payout);
+            icxToClaim.set(icxToClaim.getOrDefault(BigInteger.ZERO).add(payout));
+            icxPayable.set((Address) unstakeInfo.get(4), icxPayable.getOrDefault((Address) unstakeInfo.get(4),
+                    BigInteger.ZERO).add(payout));
+
+        }
+
+    }
+
+    @SuppressWarnings("unchecked")
```
nightowl121, line 722: Use entry set to loop through the map

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    public void unstake(Address to, BigInteger value, Address senderAddress) {
+        Context.call(sicxAddress.get(), "burn", value);
+        BigInteger amountToUnstake = (value.multiply(rate.getOrDefault(BigInteger.ZERO))).divide(Constant.DENOMINATOR);
+        Map<String, BigInteger> delegationPercent = delegationInPer(to);
```
nightowl121, line 726: Division by 100 should be done after multiplying with the amount. Also better to use HUNDRED_PERCENTAGE constant.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+        totalUnstakeAmount.set(totalUnstakeAmount.getOrDefault(BigInteger.ZERO).add(amountToUnstake));
+        for (String key : delegationPercent.keySet()) {
+            BigInteger prepPercent = delegationPercent.get(key);
+            BigInteger prepDelegations = this.prepDelegations.getOrDefault(key, BigInteger.ZERO);
```
nightowl121, line 730: prepdelegations is read again which was just set recently. Avoid multiple reads of db variables, instead pass it as parameters.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            BigInteger amountToRemove =
```
nightowl121, line 731: this db read can also be minimized.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    public void stake(BigInteger stakeValue) {
+        Context.call(Address.fromString(Constant.SYSTEM_SCORE_ADDRESS), "setStake", stakeValue);
+    }
+
+    @External
+    public void claimUnstakedICX(@Optional Address _to) {
+        if (_to == null) {
+            _to = Context.getCaller();
+        }
+        BigInteger payableIcx = icxPayable.getOrDefault(_to, BigInteger.ZERO);
```
nightowl121, line 315: there should be a check whether icxtoclaim is greater than payable icx.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+    }
+
+    @External(readonly = true)
+    public BigInteger getBlockHeightWeek() {
+        return blockHeightWeek.getOrDefault(BigInteger.ZERO);
+    }
+
+    @External(readonly = true)
+    public BigInteger getTodayRate() {
+        return rate.getOrDefault(BigInteger.ZERO);
```
nightowl121, line 144: Rate's default value is not zero.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,781 @@
...
+            List<Object> unstakeInfo = unstakingRequests.get(i);
+            BigInteger unstakeAmount = (BigInteger) unstakeInfo.get(1);
+            if (unstakeAmount.compareTo(balance) <= 0) {
+                payout = unstakeAmount;
+                linkedListDb.remove(linkedListDb.headId.getOrDefault(BigInteger.ZERO));
+            } else {
+                payout = balance;
+
+                linkedListDb.updateNode((Address) unstakeInfo.get(2), unstakeAmount.subtract(payout),
+                        (BigInteger) unstakeInfo.get(3), (Address) unstakeInfo.get(4),
```
nightowl121, line 709: Use common variable for unstakeInfo.get(4)

core-contracts/staking/src/main/java/network/balanced/score/core/utils/Constant.java
nightowl121: Changing data type to integer again requires the number to be changed back to biginteger in code in most of the cases as its being multiplied with big integer numbers. So keeping it in big integer makes sense.

core-contracts/staking/src/main/java/network/balanced/score/core/Staking.java
nightowl121: Sending default value would confuse if the value is coming from db or as default.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+    }
+
+    public Map<String, BigInteger> removePreviousDelegations(Address to) {
+        String addressStr = to.toString();
+        Map<String, BigInteger> previousDelegations = getDelegationInPercentage(to);
+        BigInteger balance = (BigInteger) Context.call(sicxAddress.get(), "balanceOf", to);
+        BigInteger icxHoldPreviously =
+                (balance.multiply(rate.getOrDefault(BigInteger.ZERO))).divide(ONE_EXA);
+        if (!previousDelegations.isEmpty()) {
+            addressDelegations.set(addressStr, "");
```
nightowl121, line 434: set null to gain steps.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+        }
+        return resetTopPreps();
+    }
+
+    @External
+    public void delegate(PrepDelegations[] _user_delegations) {
+        stakingOn();
+        Address to = Context.getCaller();
+        performChecks();
+        Map<String, BigInteger> previousDelegations = removePreviousDelegations(to);
```
nightowl121, line 464: Use same logic as transferUpdateDelegation.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+
+    public BigInteger delegateVotes(Address to, PrepDelegations[] userDelegations, BigInteger userIcxHold) {
+        BigInteger amountToStake = BigInteger.ZERO;
+        List<Address> similarPrepCheck = new ArrayList<>();
+        StringBuilder addressDelegations = new StringBuilder();
+        List<Address> allPrepAddresses = getPrepList();
+        for (PrepDelegations singlePrep : userDelegations) {
+            Address prepAddress = singlePrep._address;
+            BigInteger votesInPer = singlePrep._votes_in_per;
+            if (!allPrepAddresses.contains(prepAddress)) {
```
nightowl121, line 368: check if the user provided address is a valid prep

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+        Address to = Context.getCaller();
+        performChecks();
+        Map<String, BigInteger> previousDelegations = removePreviousDelegations(to);
+        BigInteger balance = (BigInteger) Context.call(sicxAddress.get(), "balanceOf", to);
+        BigInteger icxHoldPreviously = (balance.multiply(rate.getOrDefault(BigInteger.ZERO))).divide(ONE_EXA);
+        BigInteger totalPer = delegateVotes(to, _user_delegations, icxHoldPreviously);
+        if (totalPer.compareTo(HUNDRED_PERCENTAGE) != 0) {
+            Context.revert(Constant.TAG + ": Total delegations should be 100%.Your delegation preference is ");
+        }
+        if (!previousDelegations.isEmpty()) {
```
nightowl121, line 471: The condition to make netork effect is only when the user have some balance.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+        stakingOn();
+        if (!Context.getCaller().equals(sicxAddress.get())) {
+            Context.revert(Constant.TAG + ": The Staking contract only accepts sICX tokens." + Context.getCaller() +
+                    " or " + sicxAddress.get());
+        }
+        try {
+            String unpackedData = new String(_data);
+            JsonObject json = Json.parse(unpackedData).asObject();
+            String method = json.get("method").asString();
+            if ((json.contains("method")) && method.equals("unstake")) {
```
nightowl121, line 248: Redundant check for contains.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+                if (json.contains("user")) {
```
nightowl121, line 249: json.get("user"), check if this value is null.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+        checkForIscore();
+        checkForBalance();
+    }
+
+    public void delegations(BigInteger evenlyDistributeValue) {
+        List<Map<String, Object>> delegationList = new ArrayList<>();
+        int topPrepCount = topPreps.size();
+        BigInteger votingPowerCheck = BigInteger.ZERO;
+        for (int i = 0; i < this.topPreps.size(); i++) {
+            Map<String, Object> delegateDict = new HashMap<>();
```
nightowl121, line 515: Avoid Hashmap, use Map.of() directly.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+
+    }
+
+    @SuppressWarnings("unchecked")
+    public void unstake(Address to, BigInteger value, Address senderAddress) {
+        Context.call(sicxAddress.get(), "burn", value);
+        BigInteger amountToUnstake = (value.multiply(rate.getOrDefault(BigInteger.ZERO))).divide(ONE_EXA);
+        Map<String, BigInteger> delegationPercent = getDelegationInPercentage(to);
+        totalUnstakeAmount.set(totalUnstakeAmount.getOrDefault(BigInteger.ZERO).add(amountToUnstake));
+        for (String key : delegationPercent.keySet()) {
```
nightowl121, line 663: Use Entry set

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+            BigInteger prepPercent = delegationPercent.get(key);
+            BigInteger prepDelegations = this.prepDelegations.getOrDefault(key, BigInteger.ZERO);
+            BigInteger amountToRemove = prepPercent.multiply(amountToUnstake).divide(HUNDRED_PERCENTAGE);
+            this.prepDelegations.set(key, prepDelegations.subtract(amountToRemove));
```
nightowl121, line 667: sum of each amount to remove should be equal to amountToUnstake.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+        }
+        totalStake.set(totalStake.getOrDefault(BigInteger.ZERO).subtract(amountToUnstake));
+        delegations(resetTopPreps());
+        stake(totalStake.getOrDefault(BigInteger.ZERO));
```
nightowl121, line 671: reduce access to totalStake variable.

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+        Map<String, Object> stakeInNetwork = (Map<String, Object>) Context.call(SYSTEM_SCORE_ADDRESS, "getStake",
+                Context.getAddress());
+        Address addressToSend = to;
+        if (senderAddress != null) {
+            addressToSend = senderAddress;
+        }
+        List<Map<String, Object>> result = (List<Map<String, Object>>) stakeInNetwork.get("unstakes");
+        Map<String, Object> recentUnstakeInfo = result.get(result.size() - 1);
+        unstakeRequestList.append(to, amountToUnstake,
+                (BigInteger) recentUnstakeInfo.get("unstakeBlockHeight"),
```
nightowl121, line 681: Can we get alternative way to calculate unstake block height like add current block height with unstaking period?

core-contracts/staking/src/main/java/network/balanced/score/core/staking/Staking.java
```diff
@@ -0,0 +1,710 @@
...
+    public void checkForBalance() {
+        BigInteger balance =
+                Context.getBalance(Context.getAddress()).subtract(dailyReward.getOrDefault(BigInteger.ZERO))
+                        .subtract(icxToClaim.getOrDefault(BigInteger.ZERO));
+        if (balance.compareTo(BigInteger.ZERO) <= 0) {
+            return;
+        }
+        List<List<Object>> unstakingRequests = getUnstakeInfo();
+        for (int i = 0; i < unstakingRequests.size(); i++) {
+            if (BigInteger.valueOf(i).compareTo(unstakeBatchLimit.getOrDefault(BigInteger.ZERO)) > 0) {
```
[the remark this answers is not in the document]
nightowl121: unstake batch limit is already set as biginteger in the database from python code. Database would be required which is not desired.

AntonAndell:
From a quick glance, should we not move the interfaces folder outside to its own folder. Since they will eventually not only be used for testing but also for interscore calls. Also shouldn't you be able to implement the staking Interface in your implementation? Which i think we should do, then it makes even less sense to keep them in the testing directory.

nightowl121:
Yeah, thinking of a score-common or score-lib folder where the interfaces can remain and also the structs ( like PrepDelegations in this score) used in parameters and return types.

AntonAndell:
Sounds good to me, it also sounds like good idea to also keep the non internal structs there as you say. We could perhaps also add other utils to the same folder. So a folder for everything you might need when developing the Contracts.

```
BalancedScoreLib
    - Interfaces
    - dataTypes
    - utils
```

Not sure about names but just have everything in a package like that.

AntonAndell:
@nightowl121  ill move conversation to here.
"I would consider integration test framework as done if we could run a single command and run all the integration tests automatically. Currently the integration tests is not complete in that regards. It can be run manually though.

Once we have automation in that, we could apply CI for running integration tests in local gochain node. We could run one local gochain node and run our integration tests against it."

How do you run them in the staking contract, gradlew Staking:integrationTest? With all integration tests to do mean the integration tests for each contract? So you want a general test that integration tests for the whole smart contract environment and not in each contract?

nightowl121:
`./gradlew integrationTest` will be used in CI to run all the integration tests of different contracts in CI.

Currently running `./gradlew Staking:integrationTest` has some issues. It requires some manual intervention and updating some values. It's not properly optimized. Although all the test passes, the manual intervention has to be removed as well as it should be chain independent. These are few things that still need to be worked on for integration test.

We might as well need some plugin management similar to done in OMM
https://github.com/openmoneymarket/openmoneymarket-contracts/tree/reward-integration-test/buildSrc/src/main.

AntonAndell:
Okey cool, sounds like reasonable issues to solve. I can try to help out a bit when i have free time.

nightowl121:
Please solve the merge conflicts

AntonAndell:
@Suyog007 why is this change needed for the Balanced Client? makes test cases look horrible

Suyog007:
> makes test cases look horrible

yeah it looks horrible.I was trying to stick with the best practise.

AntonAndell:
> yeah it looks horrible.I was trying to stick with the best practise.
We could probably fix the error lines/autocomplete (which i guess is why you changed?) by either generating scoreClients like proto files instead. Or configure editors look for src files in generated under builds. Which imo is the better long term solution. Integration tests will only grow more and more complex with btp and more features. Think we need to keep them very tidy and readable.

PR #77, opened 2022-03-30
Feature/baln token

cacomixtle:


AntonAndell:
IntegrationTests are probably failing since in python we had to toggle enable snapshots, which now instead disables it when it begins as enabled

nightowl121:
> IntegrationTests are probably failing since in python we had to toggle enable snapshots, which now instead disables it when it begins as enabled

Tests are failing for Governance#executeVote
and on dividends, where there is assertion for dividends to be true but actual is false. This should be coming from the changes made on dividends where the lp fees are removed but not in tests.

nightowl121 (approved):
LGTM

AntonAndell (approved):

PR #78, opened 2022-04-01
[DEX] Translate DEX into java

gedanziger:
This has all of the dex, minus the scorelib dependent parts


AntonAndell:
Would probably be a good idea here to move to Interfaces and data structures for cross contract calls. Just to verify all the input/output. Did it to Loans and found a few potential bugs.

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1562 @@
...
+
+    @External(readonly = true)
+    public Map<String, Object> loadBalancesAtSnapshot(BigInteger _id, BigInteger _snapshot_id, BigInteger _limit, @Optional BigInteger _offset) {
+        if (_offset == null) {
+            _offset = BigInteger.ZERO;
+        }
+        Context.require(_snapshot_id.compareTo(BigInteger.ZERO) >= 0, TAG + ":  Snapshot id is equal to or greater then Zero.");
+        Context.require(_id.compareTo(BigInteger.ZERO) >= 0, TAG + ":  Pool id is equal to or greater then Zero.");
+        Context.require(_offset.compareTo(BigInteger.ZERO) >= 0, TAG + ":  Offset is equal to or greater then Zero.");
+        Map<String, Object> rv = new HashMap<>();
```
AntonAndell, line 1263: rv for ReturnValue? :smile:  Give it some describing name

AntonAndell (changes requested):
Still have a bit to go through, but for now. Most logic looks fine in a initial look, but sure a few more bugs will come up during unit testing.  But overall looks like really few mistakes translating such a huge contract (jinx inc).
Some general comments on code style:
Try to stick to a code style, there is a mix of require and reverts where requires can be used. Try to use Context.require where it makes sense. 

Alot of comments in the code, most are pretty good and much better than the python ones :smile: But some comments are just repeating what the code says, try to remove those. 

Remove extra parenthesizes for BigInteger operations.

Never use double newline to separate code and a space after nested logic ('}') is usually preferred.

score-lib/src/main/java/network/balanced/score/lib/utils/LinkedListDB.java
```diff
@@ -0,0 +1,211 @@
...
+        return new NodeDB(nodeId.toString() + name);
+    }
+
+    public void updateNode(BigInteger size, Address user, BigInteger nodeId) {
+        NodeDB node = createNodeInstance(nodeId);
+        if (node.exists()) {
+            node.setValues(size, user);
+        } else {
+            Context.revert("There is no node of the provided node id.");
+        }
```
AntonAndell, line 69:
```suggestion
       Context.require(node.exists() ,"There is no node of the provided node id.");
        node.setValues(size, user);
```

score-lib/src/main/java/network/balanced/score/lib/utils/LinkedListDB.java
```diff
@@ -0,0 +1,211 @@
...
+        if (size.equals(BigInteger.ONE)) {
+            clear();
+        } else {
+            NodeDB oldHead = getNode(headId.get());
+            BigInteger newHead = oldHead.getNext();
+            headId.set(newHead);
+            getNode(newHead).setPrev(DEFAULT_NODE_ID);
+            oldHead.delete();
+            length.set(size.subtract(BigInteger.ONE));
+        }
```
AntonAndell, line 141:
```suggestion
        if (size.equals(BigInteger.ONE)) {
            clear();
            return;
        }
 
          NodeDB oldHead = getNode(headId.get());
          BigInteger newHead = oldHead.getNext();
          headId.set(newHead);
          getNode(newHead).setPrev(DEFAULT_NODE_ID);
          oldHead.delete();
          length.set(size.subtract(BigInteger.ONE));
        
```

score-lib/src/main/java/network/balanced/score/lib/utils/LinkedListDB.java
```diff
@@ -0,0 +1,211 @@
+/*
```
AntonAndell, line 1: Use requires to remove ifs / nesting.
Remove a lot of double space. 

In general try to keep `else` to a minimum just return in the negative logic ahead of it. Positive logic in else clauses is usually considerd bad practice.

Desired flow of a function is: 

```
if bad: 
    Handle and return
if also bad
    Handle and return
    
Happy logic
```

Where in our cases many if bad can be replaced with Context.require

score-lib/src/main/java/network/balanced/score/lib/utils/Check.java
```diff
@@ -40,4 +40,8 @@ public static void isContract(Address address) {
         Context.require(address.isContract(), "Address Check: Address provided is an EOA address. A contract address " +
                 "is required.");
     }
-} 
\ No newline at end of file
+
+    public static void isOn(VarDB<Boolean> scoreAddressOn) {
```
AntonAndell, line 44: Dex checks should not be in the shared one. Move to the dex package

score-lib/src/main/java/network/balanced/score/lib/utils/EnumerableSetDB.java
```diff
@@ -0,0 +1,99 @@
...
+    public V at(int index) {
+        return entries.get(index);
+    }
+
+    public boolean contains(V value) {
+        return indexes.get(value) != null;
+    }
+
+    public Integer indexOf(V value) {
+        // returns null if value doesn't exist
```
AntonAndell, line 55:
```suggestion
```

score-lib/src/main/java/network/balanced/score/lib/utils/EnumerableSetDB.java
```diff
@@ -0,0 +1,99 @@
...
+        Integer result = indexes.get(value);
+        if (result != null) {
+            return result - 1;
+        }
+        return null;
+    }
+
+    public void add(V value) {
+        if (!contains(value)) {
+            //add new value
```
AntonAndell, line 65:
```suggestion
```

token-contracts/sicx/src/main/java/network/balanced/score/tokens/sicx/SicxImpl.java
```diff
@@ -0,0 +1,118 @@
+package network.balanced.score.tokens.sicx;
```
AntonAndell, line 1: Why is sICX token inculded in this PR? should we ignore it?

core-contracts/Dex/src/test/java/network/balanced/score/core/dex/DexImplTest.java
```diff
@@ -0,0 +1,343 @@
...
+import java.util.Map;
+import java.util.HashMap;
+
+import static network.balanced.score.lib.test.UnitTest.*;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+
+public class DexImplTest extends TestBase {
```
AntonAndell, line 31: Tests looking really well structured so far :+1: 

```suggestion
import network.balanced.score.lib.test.UnitTest;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTrue;


public class DexImplTest extends UnitTest {
```

simusud:
After updating java contracts getNamedPools seems to be reset, it is returning empty list although there are 3 named pools.

Screen Shot 2022-05-10 at 15 01 52

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1544 @@
...
+
+        matchedIndex = low.subtract(BigInteger.ONE);
+
+        return balnSnapshot.at(_id).at("values").get(matchedIndex);
+    }
+
+    @External(readonly = true)
+    public BigInteger getTotalValue(String _name, BigInteger _snapshot_id) {
+        return totalSupply(namedMarkets.get(_name));
+    }
```
Skyvell, line 1235: What is the use of this method? It's essentially the same as totalSupply. _snapshot_id is never used.

score-lib/src/main/java/network/balanced/score/lib/utils/NodeDB.java
```diff
@@ -19,8 +19,8 @@ public class NodeDB {
 
     public NodeDB(String key) {
         String name = key + NAME;
-        this.size = Context.newVarDB(name + "_value1", BigInteger.class);
-        this.user = Context.newVarDB(name + "_value2", Address.class);
+        this.size = Context.newVarDB(name + "_size", BigInteger.class);
+        this.user = Context.newVarDB(name + "_user", Address.class);
```
nightowl121, line 23: Lets not change this parameter here as this linked list db is already used by another score with this key. If dex needs another db with different keys, lets create a new one.

score-lib/src/test/java/network/balanced/score/lib/TestHelper.java
```diff
@@ -0,0 +1,13 @@
...
+import org.junit.jupiter.api.function.Executable;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class TestHelper {
+    public static void expectErrorMessage(Executable contractCall, String errorMessage) {
+        AssertionError e = Assertions.assertThrows(AssertionError.class, contractCall);
+        assertEquals(errorMessage, e.getMessage());
+    }
+}
```
nightowl121, line 13: This is already available in test-lib. Let's not duplicate the code.

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1544 @@
...
+        this.checkDistributions();
+
+        // Parse the transaction data submitted by the user
+        String unpackedData = new String(_data);
+        Context.require(!unpackedData.equals(""), "Token Fallback: Data can't be empty");
+
+        JsonObject json = Json.parse(unpackedData).asObject();
+
+        String method = json.get("method").asString();
+        JsonObject params = json.get("params").asObject();
```
AntonAndell, line 809: This fails if params are not supplied, only fetch in swap

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1544 @@
...
+
+            liquidity = (_baseValue.multiply(_quoteValue)).sqrt();
+            Context.require(liquidity.compareTo(MIN_LIQUIDITY) >= 0, TAG + ": Initial LP tokens must exceed " + MIN_LIQUIDITY.toString(10));
+            MarketAdded(id, _baseToken, _quoteToken, _baseValue, _quoteValue);
+        } else {
+            // Pool already exists, supply in the permitted order.
+            Address poolBaseAddress = poolBase.get(id);
+            Address poolQuoteAddress = poolQuote.get(id);
+
+            Context.require((poolBaseAddress == _baseToken) && (poolQuoteAddress == _quoteToken), TAG + ": Must supply " + _baseToken.toString() + " as base and " + _quoteToken.toString() + " as quote");
```
AntonAndell, line 1434: use equals instead of ==

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1544 @@
...
+    @EventLog(indexed = 3)
+    public void TransferSingle(Address _operator, Address _from, Address _to, BigInteger _id, BigInteger _value) {
+    }
+
+    @EventLog(indexed = 1)
+    public void Snapshot(BigInteger _id) {
+    }
+
+    public DexImpl(@Optional Address _governance) {
+        if (_governance != null) {
```
AntonAndell, line 161: check if governance is already set instead. Don't think we wanna rerun this code after inital deployment

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1544 @@
...
+    public Map<String, BigInteger> getBalanceAndSupply(String _name, Address _owner) {
+        if (_name.equals(SICXICX_MARKET_NAME)) {
+            Map<String, BigInteger> rewardsData = new HashMap<>();
+            rewardsData.put("_balance", balanceOf(_owner, SICXICX_POOL_ID));
+            rewardsData.put("_totalSupply", totalSupply(SICXICX_POOL_ID));
+            return rewardsData;
+        }
+        BigInteger poolId = lookupPid(_name);
+        if (poolId != null) {
+            BigInteger totalSupply = (BigInteger) Context.call(stakedlp.get(), "totalSupply", poolId);
```
AntonAndell, line 1136: Method in stakedLP is "totalStaked"

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1544 @@
...
+    public BigInteger balanceOf(Address _owner, BigInteger _id) {
+        if (_id.equals(SICXICX_POOL_ID)) {
+            BigInteger orderId = icxQueueOrderId.get(_owner);
+            if (orderId == null) {
+                return BigInteger.ZERO;
+            }
+            return icxQueue.getNode(orderId).getSize();
+        } else {
+            return balance.at(_id).get(_owner);
+        }
```
AntonAndell, line 975: We need something like this to take into account stakedLP balance before migration
```suggestion
        } else {
            BigInteger balance = this.balance.at(_id).getOrDefault(_owner, BigInteger.ZERO);
            if (getDay().compareTo(continuousRewardsDay.get()) <= 0) {
                BigInteger stakedBalance = (BigInteger) Context.call(stakedlp.get(), "balanceOf", _owner, poolId);
                balance = balance.add(stakedBalance);
            }
            
            return balance;
    }
```

AntonAndell: Investigate that what changes are needed for sicx/icx pool

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1544 @@
...
+    @External(readonly = true)
+    public BigInteger getContinuousRewardsDay() {
+        return continuousRewardsDay.get();
+    }
+
+    private boolean isLockingPool(BigInteger id) {
+        boolean continuousRewardsLaunched = continuousRewardsDay.get().compareTo(currentDay.get()) <= 0;
+        boolean restrictedPoolId = (id.compareTo(FIRST_NON_BALANCED_POOL) < 0 || id.equals(USDS_BNUSD_ID) || id.equals(IUSDT_BNUSD_ID));
+
+        return restrictedPoolId && continuousRewardsLaunched;
```
AntonAndell, line 390:
```suggestion
        return restrictedPoolId && !continuousRewardsLaunched;
```

AntonAndell: @nightowl121 Realised this wont be enough. We wanna transfer LP tokens before rewards. Perhaps change the check here to see if stakedLP address is set or just allow transfers from when this contract is deployed? 

But we do wanna restrict on icx/sicx id right?

simusud:
Some conditions not handled in:
 balanceOfAt in all pool ids : 
Screen Shot 2022-05-20 at 15 47 58

totalSupplyAt in pool id 1:
Screen Shot 2022-05-20 at 15 51 37

totalBalnAt in pool id 1 and 2:
Screen Shot 2022-05-20 at 15 52 52

Contract address on berlin test net :https://berlin.tracker.solidwallet.io/contract/cxd12d5b4afdde9199bcfd8ac39cf2f3c15cc7c376#readcontract

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -882,19 +882,21 @@ protected void _transfer(Address from, Address to, BigInteger value, BigInteger
...
+                    activeAddresses.get(id).remove(from);
+            }
         }
         TransferSingle(from, from, to, id, value);
 
         updateAccountSnapshot(from, id);
         updateAccountSnapshot(to, id);
 
-        if (to.isContract()) {
+        if ((to.isContract()) && (to.equals(stakedlp.get()))) {
```
AntonAndell, line 899: "to.equals(stakedlp.get())" why is this added?

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -882,13 +882,13 @@ protected void _transfer(Address from, Address to, BigInteger value, BigInteger
         BigInteger toBalance = balance.at(id).getOrDefault(to, BigInteger.ZERO);
         balance.at(id).set(from, fromBalance.subtract(value));
         balance.at(id).set(to, toBalance.add(value));
-        if (! to.equals(stakedlp.get())) {
+        if (!to.equals(stakedlp.get())) {
```
AntonAndell, line 885: How will unstaking be handled here?

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
Suyog007: to verify the transfer is being done to the right contract.

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
Suyog007: not sure, if i get your question correctly but unstaking will be handled on stakedLp contract.

AntonAndell: We will change active addresses when some1 unstakes, but not when they stake.  Just worried it might cause problems. The unstake call in stapedlp uses this transfer function

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
AntonAndell: Irc31 is a Standard and this breaks it. My guess other contracts like the omm LP contract expects this function to be called when they transfer tokens to their contract

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1554 @@
...
+            swapIcx(_from, _value);
+
+        } else if (method.equals("_swap")) {
+
+            // Parse the slippage sent by the user in minimumReceive.
+            // If none is sent, use the maximum.
+            JsonObject params = json.get("params").asObject();
+            BigInteger minimumReceive = BigInteger.ZERO;
+            if (params.contains("minimumReceive")) {
+                minimumReceive = BigInteger.valueOf(params.get("minimumReceive").asInt());
```
AntonAndell, line 836: Rebase on main and use convertToNumber helper to avoid errors when fetching from JSON

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1554 @@
...
+        poolTotal.at(id).set(toToken, newToToken);
+
+        // Capture details for eventlogs
+        BigInteger totalBase = isSell ? newFromToken : newToToken;
+        BigInteger totalQuote = isSell ? newToToken : newFromToken;
+
+        BigInteger sendPrice = (EXA.multiply(value)).divide(sendAmount);
+
+        // Send the trader their funds
+        Context.call(toToken, "transfer", Context.getAddress(), sendAmount, null);
```
AntonAndell, line 483: This send back to the dex contract, should it not be the sender?

Suyog007: yes that was supposed to be receiver from the param.. Great catch . Thanks

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
Suyog007: yes, but the use of activeaddresses is only on getDataBatch function of dex. so i don't think changing on activeaddresses will interrupt the stakedLp flows.

AntonAndell: Yeah agree there, as long as there are no error trying to remove a adress that does not exist and adding one that already does. 
Stake: nothing changes
Unstake: remove stakelp from active, add user to active

AntonAndell: Might be enough to just have if not to or from is stakedlp

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
nightowl121: I think should be checking if staked lp address is set or not.

Suyog007:
> Some conditions not handled in: balanceOfAt in all pool ids : Screen Shot 2022-05-20 at 15 47 58
> 
> totalSupplyAt in pool id 1: Screen Shot 2022-05-20 at 15 51 37
> 
> totalBalnAt in pool id 1 and 2: Screen Shot 2022-05-20 at 15 52 52
> 
> Contract address on berlin test net :https://berlin.tracker.solidwallet.io/contract/cxd12d5b4afdde9199bcfd8ac39cf2f3c15cc7c376#readcontract

@simusud  the changes for this issue is already pushed . this is now working right?

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -0,0 +1,1554 @@
...
+            _offset = BigInteger.ZERO;
+        }
+        Context.require(_snapshot_id.compareTo(BigInteger.ZERO) >= 0, TAG + ":  Snapshot id is equal to or greater then Zero.");
+        Context.require(_id.compareTo(BigInteger.ZERO) >= 0, TAG + ":  Pool id is equal to or greater then Zero.");
+        Context.require(_offset.compareTo(BigInteger.ZERO) >= 0, TAG + ":  Offset is equal to or greater then Zero.");
+        Map<String, Object> snapshotData = new HashMap<>();
+        for (Address addr : activeAddresses.get(_id).range(_offset, _offset.add(_limit))) {
+            BigInteger snapshotBalance = balanceOf(addr, _id);
+            BigInteger stakedBalance = (BigInteger) Context.call(stakedlp.get(), "balanceOf", addr, _id);
+            snapshotBalance = snapshotBalance.add(stakedBalance);
```
AntonAndell, line 1262: This is done in the above balanceOf so this should be removed

token-contracts/sicx/src/main/java/network/balanced/score/tokens/sicx/SicxImpl.java
nightowl121: This should be removed from this PR.

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
nightowl121: Yeah should be removed as parameter and may be use another getTotalValueAt method.

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/AbstractDex.java
```diff
@@ -0,0 +1,918 @@
...
+    void _transfer(Address from, Address to, BigInteger value, Integer id, byte[] data) {
+
+        Context.require(value.compareTo(BigInteger.ZERO) >= 0,
+                TAG + ": Transferring value cannot be less than 0.");
+
+        DictDB<Address, BigInteger> poolLpBalanceOfUser = balance.at(id);
+        BigInteger fromBalance = poolLpBalanceOfUser.getOrDefault(from, BigInteger.ZERO);
+
+        Context.require(fromBalance.compareTo(value) >= 0, TAG + ": Out of balance");
+        Context.require(!isLockingPool(id), TAG + ": Nontransferable token id");
```
simusud, line 889: We need to move this line to the beginning of the method. If the the token is non transferable we will be getting "Out of balance" revert msg.

AntonAndell:
U_SECONDS_DAY seems to be off? are they changed for integrationTest or something?

AntonAndell:
@gedanziger was this way of calculating done for any special reason?
https://github.com/balancednetwork/balanced-contracts/blob/cd185fa831de18b4d9c634689a3c6e7b559bbabe/core_contracts/dex/dex.py#L1021

int(self._pool_total[_id][_fromToken] * self._pool_total[_id][_toToken] / new_token1)
gives somewhat significant difference from:
(self._pool_total[_id][_fromToken] * self._pool_total[_id][_toToken]) // new_token1

what we have in java is more aligned with the second way of doing it.

PR #80, opened 2022-04-06
Feature/add protocol owned liquidity contract

Skyvell:
**Protocol owner liquidity contract**

Implemented according to specifications in https://github.com/balancednetwork/balanced-java-contracts/issues/21


core-contracts/Liquidity/build.gradle
```diff
@@ -28,6 +28,7 @@ dependencies {
     compileOnly Dependencies.javaeeApi
 
     implementation project(':score-lib')
+    implementation project(':test-lib')
```
nightowl121, line 31: This should be testImplementation, putting it in implementation will create issues in while optimizing the jar as the test lib uses method which are not allowed in contract.

Skyvell:
@nightowl121

I merged your latest changes to make use of score-lib and test-lib. Great stuff. I got an issue after the merge though, which I can't quite understand. When running ./gradlew optimizedJar I get the following Errors, which I find is wierd since I don't use any  of those methods. Any clues what is going on?

image

core-contracts/Liquidity/build.gradle
Skyvell: Haha talk about good timing I just tagged you with exactly this issue. That solved it. Thanks.

Skyvell:
@nightowl121 

I'm not sure how you have decided to handle testing when it comes to cross-contract calls. Which methods do you want me to write tests for, and how do you want it done? My first thought was to have all the relevant contracts deployed in the unit tests, but they are not available at this point.

Perhaps, at least test that the more critical methods that are restricted adhere to those restrictions, and mock the cross-contract calls for those methods (E.g. unstakeLPTokens, withdrawLiquidity) .

I tried testing all liquidity related methods manually, but I can not provide liquidity on the testnet dex for some reason. I think it might have built up a lot of LP reward distributions (IIRC LP functionality is locked while rewards are distributing).

nightowl121:
I have been mocking the methods and have no response from those contracts. If you think that mocking would not be suitable, it would be better to do those tests in integration test. Deploying all contracts for unit tests is not recommended, use mock or switch to integration test.

This was suggestion from anton regarding mocking of other contracts.
https://github.com/balancednetwork/balanced-java-contracts/pull/54#discussion_r837790125

core-contracts/Liquidity/src/main/java/network/balanced/score/core/liquidity/Liquidity.java
```diff
@@ -0,0 +1,213 @@
...
+    public Address getStakedLP() {
+        return stakedLP.get();
+    }
+
+    @External
+    public void supplyLiquidity(Address baseToken, Address quoteToken, BigInteger baseValue, BigInteger quoteValue) {
+        this.depositToken(baseToken, baseValue);
+        this.depositToken(quoteToken, quoteValue);
+        Context.call(this.dex.get(), "add", baseToken, quoteToken, baseValue, quoteValue);
+    }
```
Skyvell, line 119: I'm reviewing if this method without a guard could cause some vulnerability. My concern is that if anyone can supply liquidity, someone could supply liquidity in some wierd ratio, or create a new pool with some wierd ratio.

I'm looking at the dex "add method", and this is what I can make of it:

**Supplying to an existing pool**
Here it seems like the method makes sure that the tokens are supplied in the pools current ratio. If the supplier tries to supply tokens in a different ratio, the excess tokens are sent back. Conclusion: no attack vector.

**Creating a new pool**
Here it seems that a new pool can be created as long as the quote token is allowed. Here someone could potentially create some wierd ratio. Conclusion: potential attack vector.

Would be great to get a second pair of eyes on this.

nightowl121: Yes this is possible through dex. Balanced UI should not be using such unverified pools as well as other users also should be careful when they are using such pools.

Skyvell: Got it. Then we probably should have some kind of guard on this method with admin access?

Checking if the pool exists does not help since an attacker could create a pool and then supply liquidity to it. Perhaps to create arbitrage opportunity as an incentive.

BennyOptions: Hmm, I don't see this as an attack vector necessarily? Could you describe how somebody could actually earn money from this?

I think about it this way. This is an "AMM" after all. It market makes for users. If a user purposely supplies at bad prices, how can they benefit from this? It's like if somebody went to Binance and tried to place a limit order to be willing to buy ICX at $100 per token, which would be a terrible trade.

Perhaps I'm missing something, let me know

Skyvell: It sounds like we are talking about two different things. This is not the users funds; it's the protocols funds that are being supplied.

This "supplyLiquidity" method calls the "add" method in the dex, so I was looking into how the "add" works to see if there might be a problem in letting anyone call the "supplyLiquidity" method (see above how it behaves when creating a new vs supplying to an existing pool).

In scenario 2 someone could create a new pool with some ratio of his choice and profit from arbitrage between the new and existing pools?

BennyOptions: Ah yeah absolutely, that would be a huge issue haha. As a DAO, I think we can make sure that we only supply to a pool that already exists. Like we should never vote to approve a proposal that says "let's start the CFT/USDS pool".

For example, I could see this being used for supplying to the sICX/bnUSD pool, BALN / bnUSD pool, etc. Pools that already exist and have good enough liquidity

Skyvell: Unless you are fine with having only_governance for the supplyLiquidity method, we could do something like:

Introduce whiltelisted pools to the POL contract. Settable only by governance.
Then we add a check to the supplyLiquidity method, to see if the pool is whitelisted before supplying liquidity.

BennyOptions: Hmm, I feel like this would make it a bit slow in terms of executing the action autonomously, right? Wouldn't it take multiple votes to get some POL going?

Personally I'm not seeing this as a major attack vector. The DAO should just never vote to supply POL to a brand-new pool, which I don't think it would really do anyway. I see the use-case of this being sICX/bnUSD, BALN/sICX, etc. etc.

I don't really have a strong opinion though. I just don't want to complicate the existing vote process. In my head, it's currently quite simple, vote to release funds to the POL SCORE, POL SCORE claims funds with an external method call, then we call some external methods to supply the liquidity and it's good to go.

Skyvell: > Personally I'm not seeing this as a major attack vector. The DAO should just never vote to supply POL to a brand-new pool, which I don't think it would really do anyway. I see the use-case of this being sICX/bnUSD, BALN/sICX, etc. etc.

**Ans**
It all comes down to how it's used I guess, but I'll give you a concrete example of what could happend if some critera are met.

**Example:**

Lets assume that two proposals are submitted for supplying liquidity.

Proposal 1: Supply liquidity to sicx/bnusd pool.
Proposal 2: Supply liqiudity to baln/bnusd pool.

After execution of those proposal, the POL contract can claim specified amount of bnusd, baln and bnusd form the DAOfund. Now the POL contract has bnusd, sicx and baln. And anyone can call the method used to supply liquidity with these funds.

A user can now call the "supplyLiquidity" method to supply liquidity to the "sicx/baln" pool (this pool does not exist according to the balanced UI). Because this pool does not exist, the "add" method in the dex will will automatically create it, with the userspecified token ratio. The user could supply the initial ratio such that he could buy up coins at a serious discount. 


> Hmm, I feel like this would make it a bit slow in terms of executing the action autonomously, right? Wouldn't it take multiple votes to get some POL going?

**Ans:**
Not necessarily. If I remember correcly, "vote_actions", actions that will be executed, can be grouped into 
the the same proposal. 

The vote actions for a proposal could be:
   1. Give POL access to funds.
   2. Whitelist liquidity pool.

BennyOptions: Alright after thinking it through, I'd say we can put one bit of safety logic in the POL contract itself.

Requirement: POL contract should never be able to supply liquidity to a pool that does not currently exist.

Solution: When a user calls supplyLiquidity, the POL contract should first check if the pool exists. If the pool does not exist, revert.

If it is relatively simple to add this if/then check, then this would be the preferrable solution. If it will be substantial code, then let's stick with using good decision making at the voting level. Also, keep in mind that only certain tokens are allowed for quote tokens (bnUSD, sICX, USDS, IUSDC), so that also limits the attack surface

Skyvell: > Solution: When a user calls supplyLiquidity, the POL contract should first check if the pool exists. If the pool does not exist, revert.

This was my first thought as well. But any user can create a new pool, so this approach can be circumvented by:

1. Create a new pool before calling the POL contract.
2. Call the supplyLiquidity method in the POL contract.

My preferred approach would be to add the following to the POL contract:

1. A set storing whitelisted Poolids.
1. Method set whitelisted Poolids (governance only).
2. Getter for whitelisted poolids.

Then the requirement for wirthdrawing liquidity would be that the pool is whitelisted. Simple check.
Would be 15-20 lines of code maybe. No real work. The tradeoff, as you pointed out, would be one extra thing to add to be executed when the proposal passes.

> Also, keep in mind that only certain tokens are allowed for quote tokens (bnUSD, sICX, USDS, IUSDC), so that also limits the attack surface

Yeah, good point.

BennyOptions: Also please deploy the initial contract with the following pools already whitelisted:

1.) sICX/bnUSD
2.) BALN/bnUSD
3.) BALN/sICX
4.) USDS/bnUSD
5.) USDC/bnUSD

Skyvell: Got it Scott. Will do.

core-contracts/Liquidity/src/main/java/network/balanced/score/core/liquidity/Liquidity.java
```diff
@@ -0,0 +1,249 @@
...
+        return staking.get();
+    }
+
+    @External
+    public void addPoolsToWhitelist(BigInteger[] ids) {
+        only(this.governance);
+        for (BigInteger id : ids) {
+            this.whitelistedPoolIds.add(id);
+        }
+    }
```
Skyvell, line 131: @nightowl121 @AntonAndell 

Do any of you know if there is an issue with using BigInteger[] as a parameter? This parameter does not show up on Iconex contract interface as a method (the section in Iconex where you can call contracts manually). It also gives me issues when writing unit tests.

BigInteger[] idsToAdd = {BigInteger.valueOf(1), BigInteger.valueOf(2)};
liquidityScore.invoke(governanceScore, "addPoolsToWhitelist", idsToAdd);

E.g. this code gives me the following error message:

"Type BigInteger[] of the last argument to method invoke(Account, String, Object...) doesn't exactly match the vararg parameter type. Cast to Object[] to confirm the non-varargs invocation, or pass individual arguments of type Object for a varargs invocation."

Skyvell, line 131: @nightowl121 @AntonAndell

Do any of you know if there is an issue with using BigInteger[] as a parameter? This parameter does not show up on Iconex contract interface as a method (the section in Iconex where you can call contracts manually). It also gives me issues when writing unit tests.

BigInteger[] idsToAdd = {BigInteger.valueOf(1), BigInteger.valueOf(2)};
liquidityScore.invoke(governanceScore, "addPoolsToWhitelist", idsToAdd);

E.g. this code gives me the following error message:

"Type BigInteger[] of the last argument to method invoke(Account, String, Object...) doesn't exactly match the vararg parameter type. Cast to Object[] to confirm the non-varargs invocation, or pass individual arguments of type Object for a varargs invocation."

Also, how do you enter a array as a paramter when sending transactions via goloop?
I Tried a bit of everything (e.g. --param ids=[1,2]).

AntonAndell: Varargs are where a function takes a variable amount of arguments, but it behaves a bit weird in contracts. But casting to object is the solution.

```
Object idsToAdd = new BigInteger[] {BigInteger.valueOf(1), BigInteger.valueOf(2)};
liquidityScore.invoke(governanceScore, "addPoolsToWhitelist", idsToAdd);
```
Otherwise you call the function with each entry in the array as a seperate parameter. so example addPoolsToWhitelist(id1, id2, id3, id4) instead of addPoolsToWhitelist({id1, id2, id3, id4})· My guess is it has to do with either the rpc or they way parameters are represented onchain

And don't think it is possible to send "complex" parameters with the goloop cli.

Skyvell: I see, so it's unpacking the array by default when an array is passed in as an argument. Thanks for the clarification man.

Skyvell:
@nightowl121 This one is ready for review. Would be much appreciated if you could review it or assign someone to do it.

PR #81, opened 2022-04-06
Translate Rewards to java

AntonAndell:
* add related interfaces to score-lib
  * add DBHelpers to score-lib
  * add time constants to score-lib


core-contracts/Rewards/src/test/java/network/balanced/score/core/rewards/RewardsTestBase.java
```diff
@@ -0,0 +1,96 @@
...
+    protected final Account governance = Account.newScoreAccount(scoreCount++);
+
+    public MockContract<DataSourceScoreInterface> loans;
+    public MockContract<DataSourceScoreInterface> dex;
+    public MockContract<MintableScoreInterface> baln;
+    public MockContract<MintableScoreInterface> bwt;
+    public Score rewardsScore;
+
+    void setup() throws Exception {
+        dex = new MockContract<DataSourceScoreInterface>(DataSourceScoreInterface.class, sm, admin);
```
AntonAndell, line 50: @nightowl121 
You can create mocks like this and the get cleaner mocks without using string methods. Not sure on the syntax yet but it is a first draft. 
This also then provides default mocks for all function (at least void ones) So you don't need to mock all default functionality for each contract.

core-contracts/Rewards/src/test/java/network/balanced/score/core/rewards/RewardsTestBase.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+
+    public void mockBalanceAndSupply(MockContract<DataSourceScoreInterface> dataSource, String name, Address address, BigInteger balance, BigInteger supply) {
+        Map<String, BigInteger> balanceAndSupply = Map.of(
+            "_balance", balance,
+            "_totalSupply", supply
+        );
+
+        when(dataSource.mock.getBalanceAndSupply(name, address)).thenReturn(balanceAndSupply);
```
AntonAndell, line 94: @nightowl121 
Mocks can then be done like this for the methods where you need special result.

Each mockContracts also has mockContracts.acccount to do invokes and such that might be needed

core-contracts/Rewards/src/test/java/network/balanced/score/core/rewards/RewardsTestBase.java
AntonAndell: Realized now that this might clash with the bug in service Manager that does not allow to contracts of the same type.

AntonAndell:
@nightowl121 Rewards is ready for a review, if you have time or assign some1 else :slightly_smiling_face:

ZeChosenName:
https://github.com/balancednetwork/balanced-java-contracts/blob/6981236d6003943ab3ceee59322ccad2e1cd3adb/core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java#L151-L178

https://github.com/balancednetwork/balanced-contracts/blob/6a3e4d01ef43483eaf6619dd91ef582f22e56dec/core_contracts/rewards/rewards.py#L103-L108

Here the arguments should be `String` rather than using `Address` and casting it to string.

ZeChosenName:
https://github.com/balancednetwork/balanced-java-contracts/blob/6981236d6003943ab3ceee59322ccad2e1cd3adb/core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java#L538-L543
The return statement is redundant

AntonAndell:
@ZeChosenName good catch mate but seems they differ between branches
https://github.com/balancednetwork/balanced-contracts/blob/continuous-integration/core_contracts/rewards/rewards.py#L118

@ajaya-man Do you know which one we should use?

simusud:
@AntonAndell Similar issues here,  getTimeOffset is showing negative value 
Screen Shot 2022-05-10 at 14 36 14
Screen Shot 2022-05-10 at 14 38 12

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
```diff
@@ -0,0 +1,693 @@
...
+    public static final String MIGRATING_TO_CONTINUOUS = "migrating_to_continuous";
+
+    public static final VarDB<Address> governance = Context.newVarDB(GOVERNANCE, Address.class);
+    public static final VarDB<Address> admin = Context.newVarDB(ADMIN, Address.class);
+    public static final VarDB<Address> balnAddress = Context.newVarDB(BALN_ADDRESS, Address.class);
+    public static final VarDB<Address> bwtAddress = Context.newVarDB(BWT_ADDRESS, Address.class);
+    public static final VarDB<Address> reserveFund = Context.newVarDB(RESERVE_FUND, Address.class);
+    public static final VarDB<Address> daofund = Context.newVarDB(DAO_FUND, Address.class);
+    public static final VarDB<Address> stakedLp = Context.newVarDB(STAKEDLP, Address.class);
+    public static final VarDB<Integer> startTimestamp = Context.newVarDB(START_TIMESTAMP, Integer.class);
```
simusud, line 80: startTimestamp here needs to BigInteger . If you look into launchBalanced in governance there we are calling setTimeOffset with BigInteger as param.

simusud:
@AntonAndell Getting error while creating Baln market, the issue is in particular claim method in rewards throwing NullPointerException. 
 'txHash': '0x6b4d408be4899b1f0a3da00c040123daf60db627a84b76b648bc93ab9d3d7e4e',

Screen Shot 2022-05-11 at 10 02 34

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
AntonAndell: Agree that it should be bigInteger though

simusud:
> @AntonAndell Getting error while creating Baln market, the issue is with claimRewards method in rewards throwing NullPointerException. 'txHash': '0x6b4d408be4899b1f0a3da00c040123daf60db627a84b76b648bc93ab9d3d7e4e',
> 
> Screen Shot 2022-05-11 at 10 02 34

@AntonAndell is this issue fixed ? I am still getting this issue.

simusud:
> It has not, sorry i have been busy and have not had the time to find and fix it just yet

No worries, please let me know when you fix it.

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/DataSourceImpl.java
```diff
@@ -0,0 +1,268 @@
...
+            BigInteger prevHoldings = RewardsImpl.balnHoldings.getOrDefault(address.toString(), BigInteger.ZERO);
+            RewardsImpl.balnHoldings.set(address.toString(), prevHoldings.add(tokenShare));
+        }
+    
+        totalDist.set(day, remaining);
+        totalValue.set(day, shares);
+        Report(day, name, remaining, shares);
+    }
+
+    private BigInteger computeTotalWeight(BigInteger previousTotalWeight,
```
gedanziger, line 193: From the latest SCORE call, we should have the update functions in each implementation feed data always, and the updateTotalWeight / update single user functions ingest all data, and discard entries before the start of continuous rewards.

If currentTime is before the continuous rewards time, it should return before updating anything.

If lastUpdateTime is before the continuous rewards time, it should be set equal to the continous reward time.

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/DataSourceImpl.java
```diff
@@ -0,0 +1,268 @@
...
+
+        if (currentTime.compareTo(originalLastUpdateTimestamp) == 1) {
+            lastUpdateTimeUs.set(currentTime);
+        } 
+
+        return newTotal;
+    }
+
+    private BigInteger computeUserRewards(BigInteger prevUserBalance, BigInteger totalWeight, BigInteger userWeight) {
+        BigInteger detlaWeight = totalWeight.subtract(userWeight);
```
gedanziger, line 262: typo - deltaWeight

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/DataSourceImpl.java
AntonAndell: So, basically for the functions updateSingleUserData and updateTotalWeight we cant to check the day before we update weight and keep everything else the same? 

So lastUpdateTime will be set to continous rewards time on the first dataingestation?

Also just realised we dont have any checks for who is allowed to ingest data. Shouldn't that only be allowed from the datasource that is begin updated?

AntonAndell:
> > It has not, sorry i have been busy and have not had the time to find and fix it just yet
> 
> No worries, please let me know when you fix it.

I am having problems reproducing it, what setup are you using, all contracts in java?

simusud:
> > > It has not, sorry i have been busy and have not had the time to find and fix it just yet
> > 
> > 
> > No worries, please let me know when you fix it.
> 
> I am having problems reproducing it, what setup are you using, all contracts in java?

Yeah , updated java contract o top of python. 
here is the rewards address: cxd1715065020ad9bde938b419c08a4f841b4938a3
if you try calling distribute or claim method getting NullPointerException.

AntonAndell:
> Yeah , updated java contract o top of python. here is the rewards address: cxd1715065020ad9bde938b419c08a4f841b4938a3 if you try calling distribute or claim method getting NullPointerException.

Are you using continous Rewards versions of dex and loans?  The problem might be it not being backwardsCompatible

simusud:
> > Yeah , updated java contract o top of python. here is the rewards address: cxd1715065020ad9bde938b419c08a4f841b4938a3 if you try calling distribute or claim method getting NullPointerException.
> 
> Are you using continous Rewards versions of dex and loans? The problem might be it not being backwardsCompatible

No, continuous rewards is not active yet.

AntonAndell:
Yeah okey, ill look into changing the rewards contract to work with old versions of Loans and Dex. The new rewards uses some fucntions that where introduced in that update.

AntonAndell:
@simusud you can try quick now i think, Might need more changes but this at least allows calling those methods initially on my end

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
```diff
@@ -0,0 +1,704 @@
...
+    }
+
+    @External
+    public boolean distribute() {
+        BigInteger platformDay = RewardsImpl.platformDay.get();
+        BigInteger day = getDay();
+        BigInteger continuousRewardsDay = RewardsImpl.continuousRewardsDay.get();
+
+        boolean distributionRequired = false;
+        boolean continuousRewardsActive = day.compareTo(continuousRewardsDay) != -1;
```
simusud, line 331: The Null Pointer issue was here. I hadn’t set continuous day so continuousRewardsDay was null. Isn’t it better to have getOrDefault while reading continuousRewardsDay ?

AntonAndell: Yeah some default behavior should perhaps be in place

simusud: @AntonAndell 
It might be better to set some bigger value as a continuous day in the constructor. So that after deploying the value won't be null and we can set the value later as per our continuous day requirement.

simusud: I found out Night Owl has a solution for that, if we handle the null case then the above mention changes isn’t required.

AntonAndell: Yeah handled the nullcause now in claimRewards. but yeah should get some proper way of handling it than setting it to a higher value

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
```diff
@@ -0,0 +1,738 @@
...
+        return dataSources;
+    }
+    
+    @External(readonly = true)
+    public Map<String, Object> getSourceData(String _name) {
+        return DataSourceDB.get(_name).getData();
+    }
+
+    @External
+    public boolean distribute() {
```
simusud, line 327: We need to be able to distribute 100 percent of the daily distribution even if we don’t have a recipient split 100 percent .  In one of the cases, total recipient split was only 70%(although this is also an issue) due to which 30k amount was left out totalDist. And this freezes platform day .

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
```diff
@@ -0,0 +1,738 @@
...
+                RewardsImpl.platformDay.set(platformDay.add(BigInteger.ONE));
+                return false;
+            }
+        }
+
+        if (dayOfContinuousRewards || !continuousRewardsIsActive) {
+            for (int i = 0; i < DataSourceDB.size(); i++ ) {
+                String name = DataSourceDB.names.get(i);
+                DataSourceImpl dataSource = DataSourceDB.get(name);
+                if (dataSource.getDay().compareTo(day) < 0) {
```
simusud, line 382: Data source day should not be incremented unless platform day has increased. I think we need to have a check for this. 
Screen Shot 2022-05-23 at 17 00 43

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
AntonAndell: Hmm yeah, the dist should not be able to differ from 100%

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
AntonAndell: Mabye we can just compare to platfromDay instead of day perhaps

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
AntonAndell: After a lot of investigation this is because of a rewards migration done a while ago. which was probably never tested from scratch. The current python contract can be deployed as is. 
Try deploy python
then update that contract with the same contract (on_update needs to be called)
Then continue with tests

simusud: Yeah that was the issue, python contract had to be updated once to update the another db there.Though, the total recipients were 100 % it wast just getRecipientSplit method only those sources that were added to that db.
Our concern is just in any of such cases we want to have the distribution 100% among available sources.

AntonAndell: Yeah can try and fix this but this is not something we have ever handled

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/DataSourceDB.java
```diff
@@ -0,0 +1,67 @@
...
+            return;
+        }
+
+        for(int i = 0; i < names.size(); i++) {
+            if (names.get(i) == name) {
+                names.set(i, topSourceName);
+            }
+        }
+    }   
+}
```
nightowl121, line 67: Should we use .equals() here for comparison?

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
```diff
@@ -0,0 +1,738 @@
...
+    }
+
+    @External
+    public void addDataProvider(Address _source) {
+        onlyOwner();
+        dataProviders.add(_source);
+    }
+
+    @External
+    public List<Address> getDataProviders() {
```
simusud, line 517: Isn’t it supposed to be read only?

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
nightowl121: This is all coming from the issue that if total dist isn't set to zero then for next day its not minted. So if we remove the variable, the minting process for next day isn't affected.

core-contracts/Rewards/src/main/java/network/balanced/score/core/rewards/RewardsImpl.java
nightowl121: This is still not handled but platform day will still increase now regardless. Platform day will not be freezed with previous day distribution not being completed.

nightowl121:
> @ZeChosenName good catch mate but seems they differ between branches https://github.com/balancednetwork/balanced-contracts/blob/continuous-integration/core_contracts/rewards/rewards.py#L118
> 
> @ajaya-man Do you know which one we should use?

Fine with using Address but need to check on Frontend side if this will still work

nightowl121 (approved):

PR #85, opened 2022-04-15
Add the missing fallback method to accept icx from dex

nightowl121:
Accept icx from dex only and reject from other scores. Use route method to send icx and perform trade.


dbrehmer (approved):
Looks good.

PR #91, opened 2022-04-22
Fix the key name for dictdb for EnumerableSetDB to be same as python

nightowl121:
Python: {var_key}_es_indexes

Java: varKey + "_es_index" (Error)
Fixed:
Java: varKey + "_es_indexes"


dbrehmer (approved):
That'll fix it!

PR #92, opened 2022-04-25
Keep the data parameter optional in router contract

nightowl121:
Previously the token contracts were sending data as b'None' in absence of data, but in java contracts we are sending new byte[0]. This create issues in contracts which haven't handled the empty byte array.


dbrehmer (approved):
That'll do it.

PR #101, opened 2022-05-12
add temporary solution to add dual minters to bnusd

AntonAndell:


score-lib/src/main/java/network/balanced/score/lib/tokens/IRC2Burnable.java
```diff
@@ -39,7 +39,7 @@ public void burn(BigInteger _amount) {
 
     @External
     public void burnFrom(Address _account, BigInteger _amount) {
-        only(minter);
+        onlyEither(minter, minter2);
```
nightowl121, line 42: lets not change here as this is used by other contracts too, override this method in the bnusd contract and make changes there.

dbrehmer (approved):
Looks good.

PR #103, opened 2022-05-18
36 rewards integrationtest

AntonAndell:
Inital wip for automated test for rewards migration.

- rewards in python get stuck on distribute at day 2 so currently rewards are deployed as java from the start
- The amount of rewards gained are not verified
- Governance is deployed as java from start which means it currently adds the sicx/bnusd pool which wont be the case on mainet
- Rewards to dao, bwt and reserve are not tested.
- Uses timeOffset which could give unexpected results


test-lib/src/main/java/network/balanced/score/lib/test/integration/Balanced.java
```diff
@@ -64,25 +68,23 @@ public class Balanced {
...
         // delegate(adminWallet);
-        ownerClient = new BalancedClient(this, owner);
-        ownerClient.governance.createBnusdMarket(BigInteger.valueOf(210).multiply(BigInteger.TEN.pow(18)));
+        setupMarkets();
     }
-
+    
     protected void deployPrep() {
         try {
             systemScore.registerPRep(BigInteger.valueOf(2000).multiply(BigInteger.TEN.pow(18)), "test", "kokoa@example.com", "USA", "New York", "https://icon.kokoa.com", "https://icon.kokoa.com/json/details.json", "localhost:9082");
```
Suyog007, line 90: @AntonAndell  Is there any way , we can scale this up to register at least 100 preps on our local.. I tried modifying on the code by using the loop and it always says Invalid parameter .  Each of the registered preps should be called from different wallets.

AntonAndell: We probably can, but should't be part of the base deployment, would slow down a lot i imagine.

But a method that adds X preps and stores them in some map that can be would be useful i imagine. 

But i struggled a bit with this as well and not sure how the checks work. But it might need a proper url and endpoint?

AntonAndell: Yeah mabye @ajaya-man can add something here, i get invalid parameter as well but the gochain logs show nothing. even when doing it from different addresses. So we are probably missing something about preps

nightowl121: I think the config has to be updated for gochain local. There is a config to have the number of main preps and sub preps. Could you try changing that?

nightowl121: The issue was not using multiple wallets to register the prep. Only one prep can be registered by a wallet.

PR #107, opened 2022-05-19
Migrate totalDebt before continous rewards

AntonAndell:


nightowl121:
@AntonAndell Please fix the merge conflicts

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java
```diff
@@ -354,7 +350,7 @@ public boolean checkForNewDay() {
         Context.require(loansOn.get(), "Balanced Loans SCORE is not active.");
         BigInteger day = _getDay();
 
-        if (currentDay.get().compareTo(day) < 0 && day.compareTo(continuousRewardDay.get()) <= 0) {
+        if (currentDay.get().compareTo(day) < 0 && day.compareTo(continuousRewardDay.get()) < 0) {
```
nightowl121, line 353: continuous reward day should be handled here for being null.

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/positions/PositionsDB.java
```diff
@@ -233,6 +234,12 @@ public static Boolean calculateSnapshot(BigInteger day, int batchSize) {
             Position position = get(accountId);
             if (snapshotId.compareTo(BigInteger.valueOf(position.getSnaps(0))) >= 0) {
                 Standings standing = position.updateStanding(snapshotId.intValue());
+                if (!position.getDataMigrationStatus("bnUSD")) {
+                    BigInteger previousTotalDebt = LoansVariables.totalDebts.getOrDefault("bnUSD", BigInteger.ZERO);
+                    BigInteger debtAmount = position.getAssets(snapshotId.intValue(), "bnUSD");
+                    LoansVariables.totalDebts.set("bnUSD", previousTotalDebt.add(debtAmount));
+                }
+
```
nightowl121, line 242: Use BNUSD_SYMBOL constant instead of exact string.

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/positions/Position.java
```diff
@@ -143,20 +145,30 @@ public BigInteger getAssetPosition(String symbol) {
...
+     
+        if (!getDataMigrationStatus(symbol)) {
             setDataMigrationStatus(symbol, true);
         }
 
         // TODO Separate out the logic for collateral and loans
         if (arrayDbContains(AssetDB.activeAssets, symbol)) {
+            BigInteger previousTotalDebt = LoansVariables.totalDebts.getOrDefault(symbol, BigInteger.ZERO);
+            BigInteger newTotalDebt = previousTotalDebt.subtract(previousDebt).add(value);
+            LoansVariables.totalDebts.set(symbol, newTotalDebt);
```
nightowl121, line 171: Isn't the value coming here as a complete new value rather than incremental value.

nightowl121 (changes requested):

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/positions/Position.java
nightowl121: Also, the debt value should only be increased if the asset is not collateral.

AntonAndell: Active assets are just the debt so should only be debt. The value is new but it does overwrite so there can be previous debt. But should only be removed if it has already been added. Which is not the case now so that needs to be fixed

AntonAndell: So this does no longer apply? 
https://github.com/balancednetwork/balanced-contracts/blob/6a3e4d01ef43483eaf6619dd91ef582f22e56dec/core_contracts/loans/loans/assets.py#L202

there are probably a few bugs in this contract if this is false on mainnet

nightowl121:
Nevermind, you can ignore the comment. I confused with the variables.

-------- Original Message --------

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java
```diff
@@ -499,7 +498,7 @@ public void retireBadDebt(String _symbol, BigInteger _value) {
                 "Balanced.");
         Context.require(asset.isActive(), TAG + ": " + _symbol + " is not an active, borrowable asset on Balanced.");
         Context.require(assetContract.balanceOf(from).compareTo(_value) >= 0, TAG + ": Insufficient balance.");
-        Context.require(badDebt.compareTo(_value) > 0, TAG + ": No bad debt for " + _symbol);
+        Context.require(badDebt.compareTo(_value) >= 0, TAG + ": No bad debt for " + _symbol);
```
nightowl121, line 501: This avoids partial payment of bad debt. Previously in python contract we had `bad_debt <= 0` as a failure condition or `bad_debt > 0` as success condition. Now we are changing it to be `badDebt >= _value`. [L507](https://github.com/balancednetwork/balanced-java-contracts/pull/107/files#diff-a712140b8a1c3c1b804a9f575a869c523d0fcf224c44146742f7d1871a4de686R507) make sure that minimum value is chosen but now the lowest value is existing bad debt amount.

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/positions/PositionsDB.java
```diff
@@ -125,7 +125,7 @@ private static Position newPosition(Address owner) {
         newPosition.setCreated(BigInteger.valueOf(Context.getBlockTimestamp()));
         newPosition.setAddress(owner);
 
-        if (isBeforeContinuousRewardDay(snapshotIndex.intValue())) {
```
nightowl121, line 128: Lets reuse snapshot index/ current day here, that would decrease the DB read count resulting in less fees.

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java
AntonAndell: im not sure if i follow. what behavior do we want for bad debt? I see in the python contract we just checked that badDebt wasnt zero. Is that the check we want instead?

so Context.require(badDebt.compareTo(BigInteger.ZERO) > 0, "No bad debt for " + _symbol);

nightowl121: Sorry about the confusion. Yes the actual behavior should be checking with 0. Checking with value limits one condition.

Example: If the bad debt is 1000, I can't specify value as 1001. Previously it used to work by burning 1000 and leaving the 1. This requires user to precisely provide the bad debt amount rather than providing a big value and utilizing only what's required.

AntonAndell: Okey yeah i see. Thanks!

PR #109, opened 2022-05-25
Fix issue while removing item from EnumerableSetDB

nightowl121:


nightowl121:
There is no usage of remove method in main branch yet, thus wouldn't require any upgrade of the contract.

AntonAndell (approved):

PR #118, opened 2022-06-02
Remove-onlyOwner()-from-setVoteDuration

BennyOptions:
During translation, setVoteDuration was accidentally set to onlyOwner. This removes that restriction


AntonAndell:
@ajaya-man this could cause some unwanted issues at the time being right?

Should we define create a issue for being able to set vote duration when casting votes, and instead keep a variable for max and min allowed vote duration?

BennyOptions:
@AntonAndell yeah I think that's a good idea, I'll close this PR and raise the issue

PR #129, opened 2022-06-09
Integration Test framework improvements

AntonAndell:

PR #132, opened 2022-06-10
Feature/remove data as submodule

nightowl121:
Add CI on PRs too


ajaya-man (approved):
Passes the unit test.
Fails in integration which is being fixed in this https://github.com/balancednetwork/balanced-java-contracts/pull/129

PR #133, opened 2022-06-11
check if position exist before appending

AntonAndell:


core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java
```diff
@@ -103,6 +103,15 @@ public boolean getLoansOn() {
         return loansOn.get();
     }
 
+
+    @External
+    public void removeZeroPosition(int id) {
+        onlyOwner();
+        if (AssetDB.getAsset(BNUSD_SYMBOL).getBorrowers().nodeValue(id).equals(BigInteger.ZERO)){
```
nightowl121, line 110: node value should be handled for null value here.

nightowl121 (approved):

PR #138, opened 2022-06-17
Unit test for dex contract

ZeChosenName:


core-contracts/Dex/src/test/java/network/balanced/score/core/dex/DexTestCore.java
```diff
@@ -285,26 +771,26 @@ void closeMock() {
     == Tests left == 
 
     == Icx/sicx pool methods == 
-    getSicxEarnings
-    withdrawSicxEarnings
-    fallback
+    getSicxEarnings //done
```
AntonAndell, line 774: remove this section if all tests are done, otherwise remove the ones with //done

AntonAndell (approved):
Readability of test could be improved but some will soon be outdated and needs slight change anyway so i think it is fine.

PR #147, opened 2022-06-23
updated priceThreshold for rebalancing in on_install

simusud:
added default price threshold in rebalancing on_install


nightowl121 (approved):
Looks good. We start with 10% in deployment which is also the current value in mainnet. This will be helpful in testnet setup. Changing of threshold can be later done through DAO vote.

PR #150, opened 2022-06-27
added delegate method on Dex, dividends and DAO fund contracts

simusud:


AntonAndell:
Instead of adding staking to contracts just for this cause, it might be better to just fetch it from governance. So we don't need set the staking address in those contracts at all. Will be a bit more expensive but should not matter for this

simusud:
> Instead of adding staking to contracts just for this cause, it might be better to just fetch it from governance. So we don't need set the staking address in those contracts at all. Will be a bit more expensive but should not matter for this.
> 
> Also don't forget to add simple unit Tests for the contracts affected

Thanks for the review. Made all the changes

core-contracts/DAOfund/src/main/java/network/balanced/score/core/daofund/DAOfundImpl.java
```diff
@@ -103,6 +104,13 @@ public Address getLoans() {
         return loansScore.get();
     }
 
+    @External
+    public void delegate(PrepDelegations[] prepDelegations) {
+        only(governance);
+        Map<String, Address> addresses = (Map<String, Address>) Context.call(governance.get(), "getAddresses");
```
AntonAndell, line 110: Think there is a method such as 
Context.call(Address.class, governance.get(), "getContractAddress", "staking");

nightowl121 (approved):

PR #166, opened 2022-07-05
Boosted Baln Feature

nightowl121:


AntonAndell:
Calling balanceOf and totalSupply before depositing seems to crash.

PR #172, opened 2022-07-07
getPoolStatsForpair method transferred from multicall to dex

ZeChosenName:


nightowl121:
Please fix the failing tests

nightowl121:
Please remove the following line
https://github.com/balancednetwork/balanced-java-contracts/blob/53b10244fd41e78259232aa162bb1a117e3aab50/core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java#L51

Although this doesn't create any issue, but it's a redundant setter which we can avoid.

nightowl121 (approved):

PR #173, opened 2022-07-08
Add accrued fees database to fee handler

BennyOptions:
Creates a database to store accrued fees from various sources. Updates token fallback method to add all fees to the accrued fees database


core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -52,11 +55,17 @@ public class FeeHandlerImpl implements FeeHandler {
...
     private final VarDB<Address> governance = Context.newVarDB(GOVERNANCE, Address.class);
+    private final VarDB<Address> loans = Context.newVarDB(LOANS, Address.class);
+    private final VarDB<Address> dex = Context.newVarDB(DEX, Address.class);
+    private final VarDB<Address> stabilityFund = Context.newVarDB(STABILITY, Address.class);
     private final VarDB<Boolean> enabled = Context.newVarDB(ENABLED, Boolean.class);
     private final ArrayDB<Address> allowedAddress = Context.newArrayDB(ALLOWED_ADDRESS, Address.class);
     private final VarDB<Integer> nextAllowedAddressesIndex = Context.newVarDB(NEXT_ALLOWED_ADDRESS_INDEX,
             Integer.class);
     private final VarDB<Address> admin = Context.newVarDB(ADMIN_ADDRESS, Address.class);
+    private final DictDB<Address, BigInteger> swapFeesAccruedDB = Context.newDictDB(DIVIDEND_TOKENS, BigInteger.class);
```
nightowl121, line 66: Its better to use different key, since keeping same keys might mix with the other database. Although the database type is different which should avoid any issues, but its better to be safe by creating different keys.

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -52,11 +55,17 @@ public class FeeHandlerImpl implements FeeHandler {
...
+    private BigInteger loanFeesAccrued;
+    private BigInteger stabilityFundFeesAccrued;
```
nightowl121, line 68: Both of these should be stored in DB for persistence. This will work though until there is next score update.

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -82,6 +91,56 @@ public Address getGovernance() {
         return governance.get();
     }
 
+    // acceptedDividendsTokens DB is used here to add the token types of fees generated by the DEX to swapFees DB
+    @External
+    public void setSwapFeesAccruedDB() {
+        for (int i = 0; i < acceptedDividendsTokens.size(); i++) {
```
nightowl121, line 97: size should be read once and stored in a local variable otherwise the cost is increased by `number of loop` times

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -82,6 +91,56 @@ public Address getGovernance() {
...
+            Address thisToken = acceptedDividendsTokens.get(i);
+            swapFeesAccruedDB.set(thisToken, swapFeesAccruedDB.get(thisToken));
+        }
+    }
```
nightowl121, line 101: this method might not be required.

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -203,6 +262,17 @@ public void tokenFallback(Address _from, BigInteger _value, byte[] _data) {
                 transferToken(sender, getContractAddress("dividends"), getTokenBalance(sender), new byte[0]);
             }
         }
+        if (_from == loans.get()) {
+            loanFeesAccrued = loanFeesAccrued.add(_value);
+        }
+
+        if (_from == dex.get() && swapFeesAccruedDB.get(sender) != null) {
+            BigInteger accruedFees = swapFeesAccruedDB.get(sender);
```
nightowl121, line 270: Use getOrDefault while getting value from `swapFeesAccruedDB` as it can give Null Pointer Exception.

nightowl121 (changes requested):

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
AntonAndell: There is a null check above which makes sense, since all fee from dex should not be in the DB

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -82,6 +91,56 @@ public Address getGovernance() {
         return governance.get();
     }
 
+    // acceptedDividendsTokens DB is used here to add the token types of fees generated by the DEX to swapFees DB
+    @External
+    public void setSwapFeesAccruedDB() {
+        for (int i = 0; i < acceptedDividendsTokens.size(); i++) {
+            Address thisToken = acceptedDividendsTokens.get(i);
+            swapFeesAccruedDB.set(thisToken, swapFeesAccruedDB.get(thisToken));
```
AntonAndell, line 99:
```suggestion
            swapFeesAccruedDB.set(thisToken, swapFeesAccruedDB.getOrDefualt(thisToken, BigInteger.ZERO));
```

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
nightowl121: Changed the logic to store the default value and if dictdb is null then the token should not be tracked for fees.

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -203,6 +262,17 @@ public void tokenFallback(Address _from, BigInteger _value, byte[] _data) {
                 transferToken(sender, getContractAddress("dividends"), getTokenBalance(sender), new byte[0]);
             }
         }
+        if (_from == loans.get()) {
```
AntonAndell, line 270: For all these ifs, use if, else if to save some gas. Also change compares to use .equals. 
```suggestion
        if (_from.equals(loans.get())) {
```

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
nightowl121: Yeah modified the logic with new update.

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
nightowl121: Fixed for all address comparisons.

core-contracts/FeeHandler/src/main/java/network/balanced/score/core/feehandler/FeeHandlerImpl.java
```diff
@@ -198,11 +256,23 @@ public void tokenFallback(Address _from, BigInteger _value, byte[] _data) {
...
+            loanFeesAccrued.set(loanFeesAccrued.getOrDefault(BigInteger.ZERO).add(_value));
+        }
+
+        BigInteger accruedFees = swapFeesAccruedDB.get(sender);
+        if (_from.equals(dex.get()) && accruedFees != null) {
+            swapFeesAccruedDB.set(sender, accruedFees.add(_value));
+        }
+        if (_from.equals(stabilityFund.get())) {
+            stabilityFundFeesAccrued.set(stabilityFundFeesAccrued.getOrDefault(BigInteger.ZERO).add(_value));
+        }
```
AntonAndell, line 275: Since this affect basically all transactions we should optimize gas a little
```suggestion
        if (_from.equals(dex.get())) {
            BigInteger accruedFees = swapFeesAccruedDB.get(sender);
            if (accuredFees == null) {
                return;
            }
            
            swapFeesAccruedDB.set(sender, accruedFees.add(_value));
            return;
        }
        
        if (_from.equals(loans.get())) {
            loanFeesAccrued.set(loanFeesAccrued.getOrDefault(BigInteger.ZERO).add(_value));
            return;
        }
    
        if (_from.equals(stabilityFund.get())) {
            stabilityFundFeesAccrued.set(stabilityFundFeesAccrued.getOrDefault(BigInteger.ZERO).add(_value));
            return;
        }
```

nightowl121: Better to use if-else ladder or switch.

nightowl121 (approved):
LGTM

PR #183, opened 2022-07-20
Rewrite dividends calculation to be based on bbaln

ZeChosenName:


AntonAndell:
We need handle the migration here properly, the time between where staked baln/last claimed to boosted launch will be unaccounted for since your previous boosted balance was 0. But we cannot either use the old staked balance as previous since the total supplies wont add up.

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
```diff
@@ -675,6 +704,39 @@ public Map<String, BigInteger> dividendsAt(BigInteger _day) {
...
+        UserKicked(user, "user kicked".getBytes());
+    }
+
+    @External
+    public void onBalanceUpdate(Address user, BigInteger bBalnBalance) {
+        Context.require(Context.getCaller().equals(boostedBalnScore.get()), TAG + " Only BBaln contract is allowed to" +
+                " call onBalanceUpdate method.");
+        updateUserDividends(user, userBalance.getOrDefault(user, BigInteger.ZERO));
+        userBalance.set(user, bBalnBalance);
+        DividendsTracker.setBBalnTotalSupply(getBoostedTotalSupply().add(bBalnBalance));
```
AntonAndell, line 726: We also need to subtract previous here

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
```diff
@@ -107,6 +112,17 @@ public void setTimeOffset(BigInteger deltaTime) {
...
+    public void setBBalnDay(BigInteger day) {
+        onlyOwner();
+        bBalnDay.set(day);
+    }
+
+    @External(readonly = true)
+    public BigInteger getBBalnDay() {
+        return getBoostedBalnDay();
+    }
+
```
AntonAndell, line 125: Lets ignore switching day.
Lets make it so that bBaln is used directly on deployment

AntonAndell: This also will entail that you don't have to separate it into two different weights

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
```diff
@@ -675,6 +704,40 @@ public Map<String, BigInteger> dividendsAt(BigInteger _day) {
         return dividendsDist;
     }
 
+    @External
+    public void onKick(Address user) {
+        Context.require(Context.getCaller().equals(boostedBalnScore.get()), TAG + " Only BBaln contract is allowed to" +
```
AntonAndell, line 709: can use only(boostedBalnScore) here

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
```diff
@@ -675,6 +704,40 @@ public Map<String, BigInteger> dividendsAt(BigInteger _day) {
...
+                "be kicked.");
+        updateUserDividends(user, userPrevBalance);
+        userBalance.set(user, getBoostedBalnBalance(user));
+        DividendsTracker.setBBalnTotalSupply(getBoostedTotalSupply().subtract(userPrevBalance));
+        UserKicked(user, "user kicked".getBytes());
+    }
+
+    @External
+    public void onBalanceUpdate(Address user, BigInteger bBalnBalance) {
+        Context.require(Context.getCaller().equals(boostedBalnScore.get()), TAG + " Only BBaln contract is allowed to" +
```
AntonAndell, line 722: can user only(boostedBalnScore) here

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
```diff
@@ -675,6 +704,40 @@ public Map<String, BigInteger> dividendsAt(BigInteger _day) {
...
 
+    @External
+    public void onKick(Address user) {
+        Context.require(Context.getCaller().equals(boostedBalnScore.get()), TAG + " Only BBaln contract is allowed to" +
+                " call onKick method.");
+        BigInteger userPrevBalance = userBalance.getOrDefault(user, BigInteger.ZERO);
+        Context.require(!userPrevBalance.equals(BigInteger.ZERO), TAG + " " + user + " User with no balance can not " +
+                "be kicked.");
+        updateUserDividends(user, userPrevBalance);
+        userBalance.set(user, getBoostedBalnBalance(user));
```
AntonAndell, line 715: No need to fetch 0 can be assumed on a kick

core-contracts/Dividends/src/test/java/network/balanced/score/core/dividends/DividendsImplTestContinuousDividends.java
```diff
@@ -59,7 +63,7 @@ void setup() throws Exception {
 
         BigInteger day = getDay();
         dividendScore.invoke(admin, "setDividendsOnlyToStakedBalnDay", day.add(BigInteger.ONE));
-        dividendScore.invoke(owner, "setContinuousDividendsDay", day.add(BigInteger.ONE));
+//        dividendScore.invoke(owner, "setContinuousDividendsDay", day.add(BigInteger.ONE));
```
AntonAndell, line 66: remove

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
```diff
@@ -442,12 +467,12 @@ public Map<String, BigInteger> getUnclaimedDividends(Address user) {
     @External
     public void claimDividends() {
```
AntonAndell, line 468: We should also update a users boosted Balance after they have claimed. to keep it as recent as possible

AntonAndell (changes requested):

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
AntonAndell: No it can be emptied for now, but keep the function to avoid errors if called by the baln token

AntonAndell: Might have jumped the gun here it might still be needed to handle the switch properly. So people do not loose any unclaimed divs

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
```diff
@@ -555,34 +556,23 @@ public void tokenFallback(Address _from, BigInteger _value, byte[] _data) {
...
+            DividendsTracker.updateBoostedTotalWeight(token, _value.subtract(dividendsToDaofund));
+
+
+        DividendsReceivedV2(_value, getDay(), _value + " tokens received as dividends token: " + token);
+        sendToken(daoFund.get(), dividendsToDaofund, token, "Daofund dividends");
+
     }
 
     @External
     public void updateBalnStake(Address user, BigInteger prevStakedBalance, BigInteger currentTotalSupply) {
```
AntonAndell, line 575: I realized we can keep this as is instead pretty much, so we catch people unstaking to go into bBaln. So when people claim or unstake all the non boosted dividends are claimed. You can have your null check in here as well though.

AntonAndell:
Looks pretty finished now just 2 comments that needs to be addressed. Then just double check code formatting and that the tests are up to date

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
Suyog007: I assume by updating a user boosted balance here means setting the current boosted balance of the user and reducing the total supply of a bbaln again ?

AntonAndell: Yeah after they have claimed

core-contracts/Dividends/src/main/java/network/balanced/score/core/dividends/DividendsImpl.java
Suyog007: there may be a case where a user may boost their free baln token instead of unstaking a balance token and boosting it. 
In that case , while claiming their dividends they will get the dividends of previously staked baln as well as newly boosted baln , and we will be setting their previous weight to null . 
And if that same user unstakes now then he may get an issue due to null in their user weight.

AntonAndell: Yeah we will need to check if it is already null in the UpdateBalnStake and if so just ignore. So you accrue your old divs by either unstaking or claiming. And after that baln should be ignored

Suyog007: condition added in the latest commit

PR #194, opened 2022-07-28
Balanced oracle

AntonAndell:
Removed unit test since they depend on new features from the multi collateral branch

### DIFF
--- a/buildSrc/src/main/java/network/balanced/gradle/plugin/utils/NameMapping.java
+++ b/buildSrc/src/main/java/network/balanced/gradle/plugin/utils/NameMapping.java
@@ -36,7 +36,7 @@ public enum NameMapping {
     staking("Staking"),
     sicx("Sicx"),
     baln("BalancedToken"),
-    feehandler("FeeHandler");
+    feehandler("FeeHandler"),
     balancedoracle("BalancedOracle");
 
     private final String module;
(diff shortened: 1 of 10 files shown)


core-contracts/BalancedOracle/src/main/java/network/balanced/score/core/balancedoracle/EMACalculator.java
```diff
@@ -0,0 +1,71 @@
...
+import static network.balanced.score.lib.utils.Constants.EXA;
+import static network.balanced.score.lib.utils.Math.exaPow;
+
+public class EMACalculator {
+    private static final BranchDB<String, VarDB<BigInteger>> movingAverages = Context.newBranchDB(
+            "exponential_moving_averages", BigInteger.class);
+    private static final BranchDB<String, VarDB<BigInteger>> lastUpdateBlock = Context.newBranchDB(
+            "last_update_blocks", BigInteger.class);
+    private static final BranchDB<String, VarDB<BigInteger>> previousPrices = Context.newBranchDB("previous_prices",
+            BigInteger.class);
```
nightowl121, line 35: Couldn't we use just DictDB in this case, since there are just key and values??

nightowl121:
Let's add this contract in the contracts sample json and name mapping so that CI can check for its deployment status.

buildSrc/src/main/java/network/balanced/gradle/plugin/utils/NameMapping.java
```diff
@@ -37,6 +37,7 @@ public enum NameMapping {
     sicx("Sicx"),
     baln("BalancedToken"),
     feehandler("FeeHandler");
```
nightowl121, line 39: Need to add a comma here at the end instead of semicolon.

nightowl121 (approved):

PR #203, opened 2022-08-15
Feature/migrate to address manager

AntonAndell:


AntonAndell:
Contract are able to be deployed whenever and should not need to be deployed until next actual change.
Only limit is ReserveFund has to be deployed before Loans

PR #208, opened 2022-08-18
inital solution, check migration status when fetching collateral

AntonAndell:


core-contracts/Loans/src/main/java/network/balanced/score/core/loans/positions/PositionsDB.java
```diff
@@ -94,6 +95,7 @@ private static Position newPosition(Address owner) {
         newPosition.setId(id);
         newPosition.setCreated(BigInteger.valueOf(Context.getBlockTimestamp()));
         newPosition.setAddress(owner);
+        newPosition.setDataMigrationStatus(SICX_SYMBOL, true);
```
nightowl121, line 98: Should we set this as null values default to false?

AntonAndell: I think we kindof need to since new positions can't me migrated

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/positions/Position.java
```diff
@@ -101,9 +107,28 @@ public void setCollateral(String symbol, BigInteger value) {
...
     public BigInteger getCollateral(String symbol) {
+        if (symbol.equals(SICX_SYMBOL)) {
+            if (!dataMigrationStatus.at(dbKey).getOrDefault(SICX_SYMBOL, false) && 
+                collateral.at(dbKey).getOrDefault(SICX_SYMBOL, BigInteger.ZERO).equals(BigInteger.ZERO)) {
+    
+                int lastSnapIndex = snaps.at(dbKey).size() - 1;
+                int lastSnap = snaps.at(dbKey).get(lastSnapIndex);
+                BigInteger collateralAmount = assets.at(dbKey).at(lastSnap).getOrDefault(SICX_SYMBOL, BigInteger.ZERO);
+                if (collateralAmount.compareTo(BigInteger.ZERO) > 0) {
+                    setCollateral(SICX_SYMBOL, collateralAmount);
```
nightowl121, line 118: I think we can return the collateral amount from here itself rather than reading again from outside the if condition.

nightowl121 (approved):

PR #215, opened 2022-08-24
stakedLp changes

Suyog007:
- any tokens can be staked directly.
-  integration test added for stakedlp
-  governance contract changes added along with tests


AntonAndell:
Deployed to berlin
Dex -> StakedLp
Migrations works as expected
wait with governance until after voting PR

PR #219, opened 2022-08-27
Liquidate, exit early if collateral or not active

AntonAndell:


AntonAndell:
Looks good to me.

core-contracts/Loans/src/main/java/network/balanced/score/core/loans/LoansImpl.java
```diff
@@ -793,10 +793,13 @@ public void liquidate(Address _owner) {
         for (int i = 0; i < assetSymbolsCount; i++) {
             String symbol = AssetDB.assetSymbols.get(i);
             Asset asset = AssetDB.getAsset(symbol);
+            if (!asset.isActive()) {
```
dbrehmer, line 796: Probably makes sense to move the check whether the asset is collateral up to this point as well.

dbrehmer (approved):
Looks like a fix to me.

PR #220, opened 2022-08-27
set liquidation pool to zero on sweep

AntonAndell:


nightowl121:
Linked python code: https://github.com/balancednetwork/balanced-contracts/blob/6a3e4d01ef43483eaf6619dd91ef582f22e56dec/core_contracts/loans/loans/loans.py#L803

dbrehmer (approved):
Looks good.

PR #243, opened 2022-10-11
balnFees only in quote token for dex exchange

simusud:
Converted balnFees in terms of quote tokens and sent it to fee handler

### DIFF
--- a/core-contracts/Dex/src/main/java/network/balanced/score/core/dex/AbstractDex.java
+++ b/core-contracts/Dex/src/main/java/network/balanced/score/core/dex/AbstractDex.java
@@ -630,22 +630,20 @@ public abstract class AbstractDex implements Dex {
         // sent to BALN holders.
         newFromToken = newFromToken.add(lpFees);
 
+        if (isSell) {
+            BigInteger fromTokenWithoutFees = newFromToken;
+            BigInteger toTokenWithoutFees = newToToken;
+
+            newFromToken = fromTokenWithoutFees.add(balnFees);
+            newToToken = (fromTokenWithoutFees.multiply(toTokenWithoutFees)).divide(newFromToken);
+
+            balnFees = oldToToken.subtract(newToToken);
+        }
+
         // Save updated pool totals
         totalTokensInPool.set(fromToken, newFromToken);
         totalTokensInPool.set(toToken, newToToken);
 
-        if (isSell) {
-            oldFromToken = newFromToken;
-            oldToToken = newToToken;
-
-            BigInteger newFromTokenWithBalnFee = oldFromToken.add(balnFees);
-            BigInteger newToTokenAfterFeeSwap = (oldFromToken.multiply(oldToToken)).divide(newFromTokenWithBalnFee);
-
-            balnFees = oldToToken.subtract(newToTokenAfterFeeSwap);
-
-            totalTokensInPool.set(fromToken, newFromTokenWithBalnFee);
-            totalTokensInPool.set(toToken, newToTokenAfterFeeSwap);
-        }
         // Capture details for event logs
         BigInteger totalBase = isSell ? newFromToken : newToToken;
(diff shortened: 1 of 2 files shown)


AntonAndell:
Other than: https://github.com/balancednetwork/balanced-java-contracts/commit/3c2b5e3fdc8d0b30aa3207e6bae2a00886a1eae6#comments

There seems to be no tests catching that fees are now calculated differently. So also add unit tests verifying that it behaves accurately.

Be sure to also verify that amounts and such so that there are no weird rounding errors.

simusud:
> Other than: [3c2b5e3#comments](https://github.com/balancednetwork/balanced-java-contracts/commit/3c2b5e3fdc8d0b30aa3207e6bae2a00886a1eae6#comments)
> 
> There seems to be no tests catching that fees are now calculated differently. So also add unit tests verifying that it behaves accurately.
> 
> Be sure to also verify that amounts and such so that there are no weird rounding errors.

Yeah working on tests. Just wanted to verify the approach is right.

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/AbstractDex.java
```diff
@@ -633,6 +634,18 @@ void exchange(Address fromToken, Address toToken, Address sender,
...
+            oldToToken = newToToken;
+
+            BigInteger newFromTokenWithBalnFee = oldFromToken.add(balnFees);
+            BigInteger newToTokenAfterFeeSwap = (oldFromToken.multiply(oldToToken)).divide(newFromTokenWithBalnFee);
+
+            balnFees = oldToToken.subtract(newToTokenAfterFeeSwap);
+
+            totalTokensInPool.set(fromToken, newFromTokenWithBalnFee);
+            totalTokensInPool.set(toToken, newToTokenAfterFeeSwap);
+        }
```
AntonAndell, line 648:
```suggestion
        if (isSell) {
            BigInteger fromTokenWithoutFees = newFromToken;
            BigInteger toTokenWithoutFees = newToToken;

            newFromToken = fromTokenWithoutFees.add(balnFees);
            newToToken = (fromTokenWithoutFees.multiply(toTokenWithoutFees)).divide(newFromToken);

            balnFees = oldToToken.subtract(newToToken);
         }
        
         totalTokensInPool.set(fromToken, newFromToken);
         totalTokensInPool.set(toToken, newToToken);
```

Not sure if i wrote that completely correct but we can avoid the additional setter as well.

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/AbstractDex.java
```diff
@@ -600,6 +600,7 @@ void exchange(Address fromToken, Address toToken, Address sender,
         BigInteger fees = lpFees.add(balnFees);
 
         Address poolBaseToken = poolBase.get(id);
+        Address poolQuoteToken = poolQuote.get(id);
         boolean isSell = fromToken.equals(poolBaseToken);
```
AntonAndell, line 604: we can avoid the additional getter as well if we want to optimize slightly
```suggestion
        boolean isSell = fromToken.equals(poolBaseToken);
        Address poolQuoteToken = isSell ? toToken : fromToken
```

AntonAndell:
@simusud we have had some discussion on what approach is the best. The one we have here or taking fees on either ingoing or outgoing depending on which is quote token.
Personally i like this approach. It is clean and isolated from all the other logic. Looks very simple since its just a repeat of the swap math. Seems like there would be very small chance of bugs. 

However the other might be even simpler but changes the math we have currently, which has been decided that it would be okey/negligible. 

@simusud @ajaya-man  what is your stance on this. Proceed with the development done here or change to incoming/outgoing? What would you feel most comfortable with?

nightowl121 (approved):

PR #244, opened 2022-10-12
Use json object instead of hashmap to create json data bytes

nightowl121:


AntonAndell (approved):
LGTM, deployed and tested on berlin

PR #272, opened 2022-12-06
Simplify daofund asset management

AntonAndell:


core-contracts/DAOfund/src/main/java/network/balanced/score/core/daofund/DAOfundImpl.java
```diff
@@ -122,31 +113,16 @@ public Map<String, Object> getDisbursementDetail(Address _user) {
...
-            boolean requiredCondition = amountInDaofund.compareTo(asset.amount) >= 0;
-            Context.require(requiredCondition, TAG + ": Insufficient balance of asset " + asset.address.toString() +
-                    " in DAOfund");
-
-            awards.at(_recipient).set(asset.address, amountToBeClaimedByRecipient.add(asset.amount));
-            fund.set(asset.address.toString(), amountInDaofund.subtract(asset.amount));
-        }
-        return Boolean.TRUE;
+        BigInteger balance = Context.call(BigInteger.class, token, "balanceOf", Context.getAddress());
+        boolean requiredCondition = balance.compareTo(amount) >= 0;
```
nightowl121, line 121: Disburse makes an assumption that all the previous daofund amount has been claimed. We need to make sure that all amount has been claimed before creating new disburse vote proposals.

AntonAndell: Yes, I do not think we have any outstanding claims. 
But even if we did they should be small and they need to claim at any point the daofund has sufficient balance.
So taking into account very old disbursements is just unnecessary complexity.

PR #280, opened 2023-01-03
Feature/support dollar value redemptions

AntonAndell:
Release/change pricing calculations from loop to usd


AntonAndell:
Tests for loans are fixed in the loans PR:
feature/convert-loans-calculations-into-dollars

PR #290, opened 2023-01-20
Investigate and create XCall standards for Balanced

AntonAndell:
Currently annotations that supports creating and handling simple XCall methods.
Not very tested, hard to test annotations without implementation, plan is to get this working good enough to approach our first xcall implementation and improve during that process.


AntonAndell:
Moved to separate lib for now https://github.com/AntonAndell/XCall-lib

PR #296, opened 2023-01-25
Feature/improve balanced emergency shutoff and blacklist

AntonAndell:


score-lib/src/main/java/network/balanced/score/lib/utils/BalancedEmergencyHandling.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @SuppressWarnings("unchecked")
+    public void fetchBlacklist() {
+        BalancedEmergencyHandling.blacklist = (Map<String, Boolean>) Context.call(getEmergencyManager(), "getBlacklist");
+        if (BalancedEmergencyHandling.blacklist == null) {
+            BalancedEmergencyHandling.blacklist = Map.of();
+        }
+    }
+
+    public boolean isBlacklisted(String address) {
```
nightowl121, line 82: This seems to have a time complexity of O(n). While checking if a single address is blacklisted or not. It would be better to check for a single address only rather than updating the entire set.

AntonAndell: Only on deployment. Otherwise the list is not fetched at all,

nightowl121 (approved):

AntonAndell:
Deploy all contacts, order irrelevant
daofund
dex
dividends
loans
reserve
rewards
stakedLP
Staking
bBaln
staking
stability
sICX
bnUSD
Baln 

on sICX and staking:
setEmergencyManager to governance

Stability and bBaln uses Owner, so if it is owned by governance governance can disable it.
Rest uses governance as standard

on Governance:
setShutdownPrivilegeTimeLock(BigInteger days) 
addAuthorizedCallerShutdown(Address address)

core-contracts/Governance/src/main/java/network/balanced/score/core/governance/GovernanceImpl.java
```diff
@@ -319,6 +320,68 @@ public void execute(String transactions) {
         ArbitraryCallManager.executeTransactions(transactions);
     }
 
+    @External
+    public void enable() {
+        EmergencyManager.authorizeEnableAndDisable();
+        EmergencyManager.enable();
```
nightowl121, line 326: This action takes a lot of time to execute, we would need to optimize this action. Same with the disable action.

score-lib/src/main/java/network/balanced/score/lib/utils/Check.java
```diff
@@ -32,6 +32,24 @@ public static void onlyOwner() {
...
     }
 
+    public static void checkStatus() {
+        checkStatus(getGovernance());
+    }
+
+    public static void checkStatus(VarDB<Address> address) {
+        Address handler = address.get();
+        if (handler == null) {
+            return;
```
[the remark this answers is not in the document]
AntonAndell: This version of the status check is only used by sICX and Staking so i left the possibility for them to disable this feature. But it could be enforced there aswell. I do not really have a strong opinion about it.

PR #304, opened 2023-02-25
Improve loans read methods

AntonAndell:
Add method to read from linked list
re-Add null check for depositAndBorrow.


nightowl121 (approved):
Approved

PR #306, opened 2023-03-06
reintroduce readonly boolean parameter is some places to complement r…

AntonAndell:
…eadonly check


nightowl121 (approved):

nightowl121 (changes requested):
Need to update version

nightowl121 (approved):

PR #312, opened 2023-04-26
feat: implement new bbaln penalty structure

nightowl121:


AntonAndell (approved):
Looks good to me now, @nightowl121 please take a look at the changes

PR #325, opened 2023-06-09
Implement unit test for assetmanager and asset token

AntonAndell:


token-contracts/AssetToken/src/main/java/network/balanced/score/tokens/asset/AssetTokenImpl.java
```diff
@@ -73,29 +73,18 @@ public void govHubTransfer(String _from, String _to, BigInteger _value, @Optiona
...
     }
 
-    @External
-    public void burn(BigInteger _amount) {
-        burnFrom(Context.getCaller().toString(), _amount);
-    }
-
     @External
     public void burnFrom(String _account, BigInteger _amount) {
-        checkStatus();
```
nightowl121, line 83: Do we avoid checking of status now on crosschain methods?

nightowl121:
Could you also fix the integration test @AntonAndell

token-contracts/AssetToken/src/main/java/network/balanced/score/tokens/asset/AssetTokenImpl.java
AntonAndell: I removed it to avoid duplicates the method in asset manager that calls this checks the status

AntonAndell: Tried to avoid duplicated since it is a permission method, added a status check to withdraw method in asset manager

PR #341, opened 2023-10-24
feat: Allow deposits of tokens to dex via xTokenFallback

AntonAndell:


score-lib/src/main/java/network/balanced/score/lib/utils/Versions.java
```diff
@@ -33,12 +33,12 @@ public class Versions {
     public final static String DEX = "v1.0.3";
```
nightowl121, line 33: Isn't the dex being upgraded here too?

AntonAndell: Yes it does, good catch, will bump

nightowl121 (approved):

PR #344, opened 2023-10-26
Balanced crosschain update

AntonAndell:


ibrizsabin:
Didnt find any major issues and concern on review , only few refactoring and input validations where necessary and applicable.

### AssetManagerImpl

#### Deploy Asset

-   Validate symbol/ name not empty and decimals greater than zero.
    

#### WithdrawTo

-   Early exit if amount is zero.
    

  
  

### SpokeTokenImpl

#### _transfer

-   Instead of using isNative(_to) and isNative(_from) multiple times, store the results in variables and reuse them.
    
-   Instead of using Address.fromString(_to.account()) and Address.fromString(_from.account()) multiple times, store the results in variables and reuse them.
    

  

#### Burn

-   Converting `burn` to bytes every time it is called can be inefficient and unnecessary. By creating a constant byte array for `burn`, we can eliminate the need for this conversion and improve the performance of the code.
    

  

### HubTokenImpl

#### crossTransfer

-   Ensure Value in not zero or negative
    

#### xCrossTransferRevert

-   Line: 121 : Check for null value of `spokeContract` and handle it appropriately
    

#### xCrossTransfer

-   Line: 148 Define a constant for the method name `xTokenFallback`
    

  

### AssetTokenImpl

#### mintAndTransfer

-   Line: 82 Add a check to ensure `_amount` is greater than zero.
    

  

### BalancedDollarContract

#### mintWithTokenFallback

-   Line: 128 Use a constant for `tokenFallback`.
    

### DaoFund

#### Disburse

-   Ensure amount in param is greater than zero or early exit.
    

    

### LoansContract

No Issues Found.

### RouterContract

No Issues Found in current changes.

  

### XCallManager

No Issues.

nightowl121 (approved):

PR #358, opened 2024-01-08
use None instead of empty data in router

AntonAndell:


nightowl121:
Please update the version too.

nightowl121 (approved):
Approved

PR #384, opened 2024-02-29
Implement spoke xCall manager in java

AntonAndell:


spoke-contracts/SpokeXCallManager/src/main/java/network/balanced/score/spoke/xcall/manager/SpokeXCallManagerImpl.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    public boolean _verifyProtocols(String[] protocols) {
+        ProtocolConfig cfg = protocolConfig.get();
+        if (cfg.sources.length == 0) {
+            return protocols == null || protocols.length == 0;
+        }
+
+        for (String source : cfg.sources) {
+            if (!hasSource(source, protocols)) {
```
nightowl121, line 112: This can be optimized but I think can be left as todo for now due to number of protocols.

spoke-contracts/SpokeXCallManager/src/main/java/network/balanced/score/spoke/xcall/manager/SpokeXCallManagerImpl.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @External(readonly = true)
+    public String getProposedRemoval() {
+        return proposedRemoval.get();
+    }
+
+    @External
+    public void setAdmin(Address _admin) {
+        only(admin);
+        admin.set(_admin);
```
nightowl121, line 144: There can be possibility of setting an unaccessible/unauthorized admin. In that case it would have to be overridden by a contract update.

spoke-contracts/SpokeXCallManager/src/main/java/network/balanced/score/spoke/xcall/manager/SpokeXCallManagerImpl.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @External
+    public void handleCallMessage(String _from, byte[] _data, @Optional String[] _protocols) {
+        only(xCall.get());
+        Context.require(_from.equals(iconGovernance.get()), "Only ICON Governance");
+        if (!_verifyProtocols(_protocols)) {
+            ObjectReader reader = Context.newByteArrayObjectReader("RLPn", _data);
+            reader.beginList();
+            String method = reader.readString().toLowerCase();
+            Context.require(method.equals("configureprotocols"), "Invalid protocols used to deliver message");
```
nightowl121, line 171: use method name in camel case

AntonAndell: Its made into lower before compare to avoid difference in case during encoding on different chains

spoke-contracts/SpokeXCallManager/src/main/java/network/balanced/score/spoke/xcall/manager/SpokeXCallManagerImpl.java
AntonAndell: Yeah not sure how to mitigate his risk. it is a address so it is verified. Some more complex logic where one has to apply to be admin would be possible.

nightowl121 (approved):
Approved

PR #392, opened 2024-03-05
slippage control while adding liquidity on Dex contract

TRSSMG:


AntonAndell (changes requested):
Define default supply slippage as a constant along with other dex constants. 
For 10000 use global constant POINTS 
For checking default there should be a function optionalDefault in Checks library.

BigInteger default is 0 so null checks would not work in this case.
However this might be a problem if someone wants to supply with ZERO slippage. @nightowl121 Do you think we should default to Zero instead?

nightowl121:
Yeah, I think the default value should be 0%. If a user supplies 0% or defaults to 0%, both means that users wants an exact price while adding LP tokens which may or may not be possible.

AntonAndell:
Okey so no need for default slippage. default will be 0 if a optional BigInteger is omitted. So it is fine to just use that directly

AntonAndell (changes requested):
Looks good, just a few comments. Don't forget to check formatting

core-contracts/DAOfund/src/main/java/network/balanced/score/core/daofund/POLManager.java
```diff
@@ -73,7 +73,7 @@ public static void supplyLiquidity(Address baseAddress, BigInteger baseAmount, A
 
         Context.call(baseAddress, "transfer", dex, baseAmount, tokenDepositData);
         Context.call(quoteAddress, "transfer", dex, quoteAmount, tokenDepositData);
-        Context.call(dex, "add", baseAddress, quoteAddress, baseAmount, quoteAmount, true);
+        Context.call(dex, "add", baseAddress, quoteAddress, baseAmount, quoteAmount, true, getPOLSupplySlippage());
```
AntonAndell, line 76: Remove the above logic that no longer is needed, that is manually checking slippage

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/DexImpl.java
```diff
@@ -359,6 +364,8 @@ public void add(Address _baseToken, Address _quoteToken, BigInteger _baseValue,
         BigInteger poolLpAmount = poolLpTotal.getOrDefault(id, BigInteger.ZERO);
         BigInteger userLpAmount = balance.at(id).getOrDefault(user, BigInteger.ZERO);
 
+
+
```
AntonAndell, line 368:
```suggestion
```

AntonAndell (changes requested):
Looks good, just two minor changes

AntonAndell (approved):

PR #404, opened 2024-03-15
oracle protection feature on dex

TRSSMG:


AntonAndell (changes requested):

core-contracts/Dex/src/main/java/network/balanced/score/core/dex/AbstractDex.java
```diff
@@ -562,6 +610,8 @@ void exchange(Address fromToken, Address toToken, Address sender,
         // Save updated pool totals
         totalTokensInPool.set(fromToken, newFromToken);
         totalTokensInPool.set(toToken, newToToken);
+        BigInteger basePrice = fromToken.equals(poolBaseToken) ? newToToken.multiply(EXA).divide(newFromToken) : newFromToken.multiply(EXA).divide(newToToken);
+        oracleProtection(id, basePrice);
```
AntonAndell, line 615: Below this you can find endingPrice and perhaps use that instead. So we do not need to calculate it twice

AntonAndell (changes requested):
Looks good just on minor optimization

AntonAndell (approved):
LGTM

PR #410, opened 2024-03-21
feat: migrate to per asset rate limits

AntonAndell:


AntonAndell:
supposed to  not reset any limits and should be a pain free upgrade

nightowl121:
This update would bring changes to the score lib and eventually to all the contracts using this library. Do we need to do change management arising due to the change in library? Currently we are only performing change management on the actual code of the contracts.

nightowl121 (approved):
Approved

AntonAndell:
> This update would bring changes to the score lib and eventually to all the contracts using this library. Do we need to do change management arising due to the change in library? Currently we are only performing change management on the actual code of the contracts.

How do you mean? It will be applied along with other updates as we deploy them. If we want to release this separately to some contract we need to create a new PR with a new version

PR #411, opened 2024-03-21
Router: stability fund swap implementation

TRSSMG:


nightowl121 (approved):
Approved

AntonAndell (changes requested):
Route v2 also need to be a option in tokenFallback

core-contracts/Router/src/main/java/network/balanced/score/core/router/RouterImpl.java
```diff
@@ -276,23 +313,39 @@ public void xTokenFallback(String _from, BigInteger _value, byte[] _data) {
...
-        Address[] path = new Address[pathArray.size()];
-
-        for (int i = 0; i < pathArray.size(); i++) {
-            JsonValue addressJsonValue = pathArray.get(i);
-            if (addressJsonValue == null || addressJsonValue.toString().equals("null")) {
-                path[i] = null;
-            } else {
-                path[i] = Address.fromString(addressJsonValue.asString());
+        List<RouteAction> actions = new ArrayList<>();
+        if (!params.contains("isStableSwap")) {
```
AntonAndell, line 317: Instead use the method field here.
Switch on method and default revert.
"_swap"
and maybe 
"_swapv2"

No need for the extra parameter

core-contracts/Router/src/main/java/network/balanced/score/core/router/RouterImpl.java
```diff
@@ -276,23 +319,42 @@ public void xTokenFallback(String _from, BigInteger _value, byte[] _data) {
...
         }
 
         Address fromToken = Context.getCaller();
-        route(receiver, fromToken, path, minimumReceive);
+        route(receiver, fromToken, actions, minimumReceive);
+    }
+
+    private void tokenFallbackV2(String _from, byte[] data) {
+        RouteData routeData = RouteData.fromBytes(data);
+        Context.require(routeData.method.contains("_swapV2"), TAG + ": Fallback directly not allowed.");
```
AntonAndell, line 342: Do we need a method at all here?

AntonAndell (commented):
Looks good, just one minor comment for now. WIll take another look tomorrow

core-contracts/Router/src/main/java/network/balanced/score/core/router/RouterImpl.java
TRSSMG: ya, I think, separation is better

AntonAndell: How do you mean separation? What does the extra string "_swapv2" add information wise?

I can perhaps see if we want to add more methods in the future. But we already have actions for that

core-contracts/Router/src/main/java/network/balanced/score/core/router/RouterImpl.java
```diff
@@ -283,17 +287,41 @@ public void tokenFallback(Address _from, BigInteger _value, byte[] _data) {
 
     @External
     public void xTokenFallback(String _from, BigInteger _value, byte[] _data) {
-        // Receive token transfers from Balanced DEX and staking while in mid-route
-        if (_from.equals(getDex().toString()) || _from.equals(MINT_ADDRESS.toString())) {
+        if (inRoute) {
             return;
```
TRSSMG, line 291: this also should be revert, other wise the token transfer will be success, while underlying reason for transfer will not execute

AntonAndell: No, this we need to accept deposits during the route. Only if someone tries to reenter(which they should not be able to) the will simply lose their deposit

TRSSMG: ok, got it

AntonAndell (approved):

PR #414, opened 2024-04-23
feat: Add logic to override deposits from chains

AntonAndell:


core-contracts/AssetManager/src/main/java/network/balanced/score/core/asset/manager/AssetManagerImpl.java
```diff
@@ -73,9 +73,6 @@ public AssetManagerImpl(Address _governance, byte[] tokenBytes) {
         }
 
         currentVersion.set(Versions.BALANCED_ASSET_MANAGER);
-        if (dataMigrated.get() == null) {
```
TRSSMG, line 76: can we delete this dataMigrated variable now? and remove the storage definition later as this variable has no use case now ?

AntonAndell: You are right, we should remove DB aswell

TRSSMG (approved):

PR #417, opened 2024-05-09
feat: Use protocols when estimating daofund xCall fees 

AntonAndell:
Also allow unchecked rollbacks

### DIFF
--- a/core-contracts/DAOfund/src/main/java/network/balanced/score/core/daofund/DAOfundImpl.java
+++ b/core-contracts/DAOfund/src/main/java/network/balanced/score/core/daofund/DAOfundImpl.java
@@ -18,10 +18,12 @@ package network.balanced.score.core.daofund;
 
 import network.balanced.score.lib.interfaces.DAOfund;
 import network.balanced.score.lib.structs.PrepDelegations;
+import network.balanced.score.lib.structs.ProtocolConfig;
 import network.balanced.score.lib.utils.BalancedAddressManager;
 import network.balanced.score.lib.utils.EnumerableSetDB;
 import network.balanced.score.lib.utils.Names;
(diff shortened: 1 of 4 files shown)


core-contracts/DAOfund/src/main/java/network/balanced/score/core/daofund/DAOfundImpl.java
```diff
@@ -246,7 +247,8 @@ public boolean getXCallFeePermission(Address contract, String net) {
     public BigInteger claimXCallFee(String net, boolean response) {
         Address contract = Context.getCaller();
         Context.require(xCallFeePermissions.at(contract).getOrDefault(net, false), contract + " is not allowed to use fees from daofund");
-        BigInteger fee = Context.call(BigInteger.class, BalancedAddressManager.getXCall(), "getFee", net, response);
+        Map<String, String[]> protocol = XCallUtils.getProtocols(net);
+        BigInteger fee = Context.call(BigInteger.class, BalancedAddressManager.getXCall(), "getFee", net, response, protocol.get("sources"));
```
TRSSMG, line 251: I prefer to use the string constant for the repeating strings, or even a single string, as strings in the code are error prone!

TRSSMG (approved):

core-contracts/DAOfund/src/main/java/network/balanced/score/core/daofund/DAOfundImpl.java
AntonAndell: Yeah, this is why we introduced the unit testing framework where the strings used for methods are checked against the interface. Adding strings for every method might cause problems as well as it need to be maintained separately. But for the map key i can see adding a String constant

TRSSMG (approved):

PR #418, opened 2024-05-09
feat: Add option to link together tokens with different decimals

AntonAndell:


mouhand999 (approved):

core-contracts/AssetManager/src/main/java/network/balanced/score/core/asset/manager/AssetManagerImpl.java
```diff
@@ -45,6 +46,7 @@ public class AssetManagerImpl implements AssetManager {
     public static final String DATA_MIGRATED = "data_migrated";
```
TRSSMG, line 46: DATA_MIGRATED can be removed now

core-contracts/AssetManager/src/main/java/network/balanced/score/core/asset/manager/AssetManagerImpl.java
```diff
@@ -316,6 +325,27 @@ private void _withdrawTo(Address asset, String from, String to, BigInteger amoun
...
+        return translateDecimals(token, amount, 1);
+    }
+
+    private BigInteger translateIncomingDecimals(String token, BigInteger amount) {
+        return translateDecimals(token, amount, -1);
+    }
+
+    private BigInteger translateDecimals(String token, BigInteger amount, int sign) {
+        BigInteger translation = decimalTransformations.get(token);
+        if (translation == null || translation.equals(BigInteger.ZERO)) {
```
TRSSMG, line 338: translation will never be equal to zero, we not not set the translation for ZERO case

TRSSMG:
two more things: 
- When removeToken we can remove the decimalTransformations too, I think
- Don’t we have to implement translateIncomingDecimals for method withdrawRollback?

AntonAndell:
> two more things:
> 
> * Don’t we have to implement translateIncomingDecimals for method withdrawRollback?

For the rollback i with intention did put the actual token decimals on ICON as the amount so we wouldn't need to. Please doublecheck but i think that should be covered

TRSSMG:
all good

TRSSMG (approved):

PR #455, opened 2024-10-04
feat: sicx/balan token crosschain

Itshyphen:
### Commit Message

```bash
type: commit message
```

see the [guidelines](https://github.com/icon-project/community/blob/main/guidelines/technical/software-development-guidelines.md#commit-messages) for commit messages.

### Changelog Entry

```bash
version: <log entry>
```

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have documented my code in accordance with the [documentation guidelines](https://github.com/icon-project/community/blob/main/guidelines/technical/software-development-guidelines.md#documentation)
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have run the unit tests
- [ ] I only have one commit (if not, squash them into one commit).
- [ ] I have a descriptive commit message that adheres to the [commit message guidelines](https://github.com/icon-project/community/blob/main/guidelines/technical/software-development-guidelines.md#commit-messages)

> Please review the [CONTRIBUTING.md](/CONTRIBUTING.md) file for detailed contributing guidelines.

### DIFF
--- a/score-lib/src/main/java/network/balanced/score/lib/utils/Versions.java
+++ b/score-lib/src/main/java/network/balanced/score/lib/utils/Versions.java
@@ -17,11 +17,11 @@
 package network.balanced.score.lib.utils;
 
 public class Versions {
-    public final static String BALN = "v1.0.2";
+    public final static String BALN = "v1.1.2";
     public final static String DIVIDENDS = "v1.0.0";
     public final static String LOANS = "v1.2.0";
     public final static String RESERVE = "v1.0.0";
-    public final static String SICX = "v1.0.1";
+    public final static String SICX = "v1.1.1";
     public final static String STAKING = "v1.0.1";
     public final static String WORKERTOKEN = "v1.0.0";
     public final static String BNUSD = "v1.1.1";
--- a/score-lib/src/main/java/network/balanced/score/lib/utils/Versions.java
+++ b/score-lib/src/main/java/network/balanced/score/lib/utils/Versions.java
@@ -17,11 +17,11 @@
 package network.balanced.score.lib.utils;
 
 public class Versions {
-    public final static String BALN = "v1.1.2";
+    public final static String BALN = "v1.1.0";
     public final static String DIVIDENDS = "v1.0.0";
     public final static String LOANS = "v1.2.0";
     public final static String RESERVE = "v1.0.0";
-    public final static String SICX = "v1.1.1";
+    public final static String SICX = "v1.1.0";
     public final static String STAKING = "v1.0.1";
     public final static String WORKERTOKEN = "v1.0.0";
     public final static String BNUSD = "v1.1.1";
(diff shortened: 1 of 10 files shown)


token-contracts/BalancedToken/src/main/java/network/balanced/score/tokens/balancedtoken/BalancedTokenImpl.java
```diff
@@ -647,17 +628,28 @@ public BigInteger totalStakedBalanceOfAt(BigInteger _day) {
...
+
+    @External
+    public void setMinter(Address _address) {
+        onlyOwner();
+        minter.set(_address);
+    }
+
+    @External(readonly = true)
+    public Address getMinter() {
+        return minter.get();
```
AntonAndell, line 653: How come this is added?

AntonAndell (changes requested):
Generally i think this looks correct.
We should add this logic as well for now. So we can do bridging from any chain to any chain.  
    @Override
    public BigInteger getHopFee(String net) {
        if (!canWithdraw(net)) {
            return BigInteger.ONE.negate();
        }
        return Context.call(BigInteger.class, getDaofund(), "claimXCallFee", net, false);
    }

    private boolean canWithdraw(String net) {
        return Context.call(Boolean.class, getDaofund(), "getXCallFeePermission", Context.getAddress(), net);
    }

token-contracts/BalancedToken/src/main/java/network/balanced/score/tokens/balancedtoken/BalancedTokenImpl.java
gcranju: This was in IRC2Mintable, and we were using 'minter' to mint and burn those tokens, so need to add there

Itshyphen:
> Generally i think this looks correct. We should add this logic as well for now. So we can do bridging from any chain to any chain.
> 
> ```
>     @Override
>     public BigInteger getHopFee(String net) {
>         if (!canWithdraw(net)) {
>             return BigInteger.ONE.negate();
>         }
>         return Context.call(BigInteger.class, getDaofund(), "claimXCallFee", net, false);
>     }
> 
>     private boolean canWithdraw(String net) {
>         return Context.call(Boolean.class, getDaofund(), "getXCallFeePermission", Context.getAddress(), net);
>     }
> ```

yeahh, done

score-lib/src/main/java/network/balanced/score/lib/utils/Versions.java
```diff
@@ -17,11 +17,11 @@
 package network.balanced.score.lib.utils;
 
 public class Versions {
-    public final static String BALN = "v1.0.2";
+    public final static String BALN = "v1.1.2";
```
AntonAndell, line 20: change versions to 1.1.0. Since we bumped a whole version.
Otherwise we can continue with launch

AntonAndell (approved):

PR #478, opened 2025-07-07
feat: add logic to close the balanced DAO

AntonAndell:


nightowl121:
Could you describe what is being closed after the Balanced has been shut down?
Like rewards, distribution of baln?

AntonAndell:
> Could you describe what is being closed after the Balanced has been shut down? Like rewards, distribution of baln?
This i the plan.

PreAction
Upgrade Governance to add action to disable governance and baln:

First action set
bBaln: Upgrade -> withdrawEarly now has no penalty
Fee handler: use current disable. -> Means fee will be pooled in feehandler for now. And we can decide later what we want to do with them. (Also stopping burn) (Should now be SODAX allocated funds)
Rewards: Upgrade -> diables sets emissions to 0
Loans : upgrade-< disabled sends all interest to fee handler
Trickler: Upgrade to stop distribution and make all claimable

Second action set
Loans -> configure debt ceilings to be 0 (no more loans)
We are initially keeping the stability fund in the background so that liquidations can take place.

Third action set
Remove stability fund fees.

AntonAndell:
With shutdown i mean shutdown of the DAO

nightowl121 (approved):

